PR #26, opened 2026-05-01
fix(security): wire WORKSPACE_SECRETS_KEY through prod compose; fail-closed in prod

matescb:
## Summary

- Closes the operational hole the v2 teardown flagged in `INFRA2-004` + `SEC2-002` (both Critical) — `WORKSPACE_SECRETS_KEY` was not passed into the backend container, and `core/secrets.py` shipped a hard-coded Fernet key in plaintext as the fallback. Net effect: prod was encrypting workspace credentials under a key committed to the public repo.
- Adds a Pydantic `model_validator` that fails closed at import when `APP_ENV=="prod"` and the key is empty.
- Replaces the committed dev fallback with a per-process random Fernet key (round-trips work in-process; new process = new key, fine for dev).
- Bonus: ships `docs/teardown/v1-status.md` — a finding-by-finding cross-reference of the v1 review (`review-2026-04-30/`) against the v2 teardown (`docs/teardown/`) and the 25 PRs that shipped between them. Headline: 14/22 v1 CRITs resolved, 2 partial, 3 open, 3 deferred (this PR closes the partial that mattered most).

## ⚠️ Pre-merge ops step (REQUIRED — deploy will crash-loop otherwise)

The post-merge `docker compose up` will fail closed if `WORKSPACE_SECRETS_KEY` is not set in `/srv/stockmanager/.env.prod` because the new validator raises `ValidationError` at import. Set it BEFORE merging:

\`\`\`bash
ssh deploy@parts.matescb.cz
cd /srv/stockmanager
python3 -c \"from cryptography.fernet import Fernet; print(Fernet.generate_key().decode())\"
# paste output as WORKSPACE_SECRETS_KEY=<key> in /srv/stockmanager/.env.prod
\`\`\`

Escrow the key alongside `SESSION_SECRET` — losing it makes every encrypted credential unrecoverable.

## Post-merge ops step

Rotate every workspace's stored Mouser / DigiKey API keys and Scandit license — they were written under the dev key during the PR #25 → this PR window. (User confirmed no real prod data exists, so this is a precaution rather than a real cred-rotation.)

## Test plan

- [ ] CI green (244+ backend tests + new fail-closed regression tests)
- [ ] `WORKSPACE_SECRETS_KEY` set in `/srv/stockmanager/.env.prod` before merge
- [ ] Post-deploy: `curl https://parts.matescb.cz/api/health` returns 200
- [ ] Post-deploy: log into the app, navigate to Workspace settings, verify provider-key UI doesn't error
- [ ] Post-deploy: workspaces with provider creds either round-trip cleanly (if encrypted under the new key already, e.g. fresh data) or surface a 500 on read (if encrypted under the old dev key) — clear creds + re-enter on those rows

### DIFF
--- a/backend/app/core/config.py
+++ b/backend/app/core/config.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from functools import lru_cache
-from pydantic import Field
+from pydantic import Field, model_validator
 from pydantic_settings import BaseSettings, SettingsConfigDict
 
 
@@ -32,9 +32,14 @@ class Settings(BaseSettings):
     SENTRY_TUNNEL_MAX_BYTES: int = 200 * 1024
     # Fernet key (urlsafe-base64-encoded 32 bytes) for encrypting
     # workspace-level secrets at rest: parts_provider_api_key,
-    # parts_provider_api_secret, scanner_license_key. Empty → falls
-    # back to a dev-only default (see app/core/secrets.py) with a
-    # warning logged on first use. Generate with:
+    # parts_provider_api_secret, scanner_license_key.
+    #
+    # In prod the model validator below rejects an empty value — a
+    # missing key fails the import loud rather than silently encrypting
+    # under a fallback. In dev `app/core/secrets.py` generates a
+    # per-process ephemeral key, so local runs work without a `.env`.
+    #
+    # Generate with:
     #   python -c "from cryptography.fernet import Fernet; print(Fernet.generate_key().decode())"
     WORKSPACE_SECRETS_KEY: str = ""
     CORS_ORIGINS: str = "http://localhost:5173"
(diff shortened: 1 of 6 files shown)


matescb (commented):
## Approve

**Closes:** `SEC-001`, `SEC2-002`, `INFRA2-004`. Wires `WORKSPACE_SECRETS_KEY` into prod compose; fails startup closed when it's empty in prod; replaces the committed dev fallback key with a per-process ephemeral key.

### What the patch does
1. Adds `model_validator(mode="after")` to `Settings` — raises `ValueError` if `APP_ENV == "prod"` and `WORKSPACE_SECRETS_KEY` is empty.
2. Removes the committed `_DEV_DEFAULT_KEY = b"OXmO1Y_..."` from `secrets.py`. In dev with the env var unset, mints `Fernet.generate_key()` per process (cached via `lru_cache`).
3. Adds `WORKSPACE_SECRETS_KEY: ${WORKSPACE_SECRETS_KEY}` to prod compose backend env.
4. Updates `.env.prod.example` documentation.
5. Adds three tests: prod-empty raises, prod-valid passes, dev-empty allowed.
6. Bonus: `docs/teardown/v1-status.md` reconciling the v1 review against the v2 teardown.

### Strengths
- ✅ **Right place to fail**: `model_validator(mode="after")` runs at every `Settings()` construction. Combined with `lru_cache` on `settings()` and the import-time eval, a missing prod key crashes uvicorn before it accepts a request — exactly the loud failure the v2 teardown asked for.
- ✅ **Removing the committed `_DEV_DEFAULT_KEY` is the actual security improvement**, not just the prod gate. Anyone with the source could decrypt any dev/staging dump until this PR. Now they can't.
- ✅ **Per-process ephemeral key in dev** is a clean trade-off: zero-config local runs still work; encrypted credentials don't survive a restart. Comment is upfront about the gotcha.
- ✅ Tests cover all three states. `monkeypatch.setenv` + direct `Settings()` construction correctly bypasses the `lru_cache` to exercise the validator.

### Notes (non-blocking)

1. **Validator doesn't check Fernet key shape** — `WORKSPACE_SECRETS_KEY="garbage"` passes the `bool()` check but `Fernet(b"garbage")` raises later, on first encrypt/decrypt. Tighter version: try `Fernet(self.WORKSPACE_SECRETS_KEY.encode("ascii"))` inside the validator and raise on `ValueError`/`binascii.Error`. Worth a follow-up issue, not a blocker for this PR.
2. **This PR is necessary but not sufficient on its own**: with no post-deploy health gate (`INFRA2-002`), a `.env.prod` that forgets the key crashes uvicorn into a restart loop with no alarm. PR #29 closes that loop — pair them in time.
3. **No test that the dev path doesn't accidentally generate a *new* key per call** — the `lru_cache(maxsize=1)` on `_fernet()` is what makes the ephemeral key stable for a process's lifetime. A test like `assert _fernet() is _fernet()` would pin the cache decorator. Trivial, future-proofing.

### Ops step (call-out)

This PR's branch doc explicitly flags: **set `WORKSPACE_SECRETS_KEY` in `/srv/stockmanager/.env.prod` before merge or backend will crash-loop**. Don't merge without confirming the env var is in place on the VPS.

PR #27, opened 2026-05-01
fix(security): default-deny Sentry request bodies + drop frame locals (SEC2-005)

matescb:
## Summary

- Closes the v2 teardown's `SEC2-005` (High) — the existing `before_send` scrubber stripped request bodies only on `/api/workspaces`. Every other secret-bearing route (signup/login plaintext password, invitation accept token, parts-provider API keys, bulk-import) leaked on a 5xx.
- Replaces the URL allow-list with a **method-based default-deny**: `request.data` is dropped on every non-GET. New routes are safe by construction.
- Sets `include_local_variables=False` so frame-local variables (e.g. `payload.password`, `decrypt(...)` returns) are not shipped with stack traces.
- Mirrors both changes on the frontend SDK in `web/src/instrument.ts`.
- Header scrub (Cookie / Authorization / X-Workspace-Id) is unchanged.

## Test plan

- [ ] CI green — new `backend/tests/test_sentry_scrubber.py` (13 cases) covers: body strip on POST/PATCH/DELETE/PUT (incl. signup, login, invitation accept, provider lookup, workspace patch); body kept on GET; header scrub on every method; defensive shapes (no request, no method, POST without body).
- [ ] Manual post-deploy: trigger a deliberate 5xx (e.g. invalid signup payload) and verify in Sentry that `request.data` is absent and `body_redacted: true` is set.
- [ ] Manual post-deploy: trigger an exception in a backend handler that has secrets in scope (e.g. PATCH workspace) and verify `frames[*].vars` is empty.

No ops step required — pure code change. This PR is independent of #26 (the `WORKSPACE_SECRETS_KEY` wiring); both can land in either order.

### DIFF
--- a/backend/app/main.py
+++ b/backend/app/main.py
@@ -14,17 +14,28 @@ from app.core.logging import configure_logging
 configure_logging()
 
 
-# Top-level (testable) Sentry event scrubber. Runs as `before_send` so the
-# event is mutated before it leaves the worker. We have to scrub here
-# rather than rely on Sentry's defaults because:
-# - PATCH /api/workspaces/current carries the workspace's plaintext
-#   provider API keys + scanner license key in the body. A 5xx during
-#   that PATCH would otherwise ship those credentials to Sentry.
-# - POST /api/workspaces/{id}/switch carries the target workspace_id —
-#   tenant-identifying, low value to triage.
-# - Cookie / Authorization / X-Workspace-Id headers are tenant- or
-#   session-identifying. Sentry redacts `Cookie` by default, but we
-#   strip explicitly so we don't depend on the SDK's redaction list.
+# Top-level (testable) Sentry event scrubber. Runs as `before_send` so
+# the event is mutated before it leaves the worker.
+#
+# Default-deny `request.data` on any non-GET method. The previous narrow
+# allow-list ("only /api/workspaces") was identified by the 2026-05-01 v2
+# teardown (SEC2-005) as leaking on every other route that handles a
+# secret-bearing body:
+#   - POST /api/auth/signup, /login → plaintext password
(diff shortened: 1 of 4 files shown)


matescb (commented):
## Approve

**Closes:** `SEC-002` / `SEC2-005`. Fixes Sentry leaking credential-bearing request bodies on every non-`/api/workspaces` route.

### What the patch does
- Replaces the URL allow-list (`/api/workspaces` only) with a method-based default-deny: every non-GET strips `request.data` on both backend (`_scrub_event`) and frontend (`instrument.ts`).
- Sets `include_local_variables=False` on the SDK init — drops frame locals from event payloads (closes a leak surface I called out: `payload.password`, decrypt return values, Pydantic kwargs).
- 12 new tests in `test_sentry_scrubber.py` covering signup, login, invitation accept, provider lookup, workspaces PATCH, DELETE, PUT, GET-keeps-body, header stripping, no-request, no-method, no-data edge cases.

### Strengths
- ✅ Backend and frontend scrubbers are symmetric — same posture, same comment, mirrored conditional. No drift risk.
- ✅ `method and method != "GET"` guards the missing-method case correctly (don't strip if we don't know).
- ✅ Test for `body_redacted` flag presence/absence is correct (don't lie about scrubbing if there was no body).
- ✅ `include_local_variables=False` is the unsung win — it kills a whole class of leak vectors that scrubbing `request.data` alone wouldn't catch.

### Notes (non-blocking)
- The backend comment ("There is no read-only POST in the API today") is a future-leaking statement. Acceptable since the posture is explicit, but worth re-reading the next time someone adds a search-via-POST endpoint and wonders why their query body isn't in Sentry.
- HEAD/OPTIONS also strip body. Both are bodyless in practice — no behaviour difference.

PR #28, opened 2026-05-01
fix(stock): add_stock/receive/release lock + service-trigger NULL bucket alignment

matescb:
## Summary

Closes the three gaps that PR #11 (the stock TOCTOU fix from 2026-05-01) left open, all flagged by the v2 teardown:

- **BE2-001** (Critical) — `add_stock`, `receive`, and the `output_lot` path inside `consume` now take the per-(workspace, part) advisory lock. The original PR's "positive deltas can't go negative" reasoning ignored that producer-side invariant reads (default-storage-mandatory, single_part_only, storage-archived) executed without the lock, surfacing as 500s from the 0013 trigger instead of clean 4xx.
- **BE2-008** (High) — `apply_reservations` and `release_reservations` mutate the reserved ledger; both now aggregate the part_ids they touch, sort by UUID string, and acquire the lock in deterministic order before any read. `consume` bundles all required locks (BOM ∪ lines ∪ output) up front so subsequent inner calls re-acquire as a no-op — eliminates AB/BA deadlocks between concurrent consumes on overlapping BOMs.
- **DB-002 / BE-002** (High) — `current_quantity` gains a `bucket_match: bool` parameter. When True, `None` means "the SQL NULL bucket specifically" (matches the 0013 trigger's `IS NOT DISTINCT FROM` grouping); when False (default), `None` means "don't filter" (preserves global aggregates for reports). The mutate paths (`remove`, `move`, `adjust`, `consume`) pass True so a remove with NULL coordinates against stock in a non-NULL bucket now raises a clean `StockError` instead of falling through to a check_violation 500.

New helper `lock_parts_for_stock_write(part_ids=...)` is exported from `domain/stock/service.py` and used by the multi-part call sites.

## Test plan

- [ ] CI green (244+ existing tests + new `test_stock_null_bucket.py`).
- [ ] New regression tests cover: NULL-lot remove against non-NULL stock raises 4xx; NULL-storage same; explicit-bucket happy path still works; `total_for_part` still aggregates globally.
- [ ] Existing tests using `_add_stock(... no lot)` + consume `(... no lot_id)` pattern still pass — the stock lands at lot=NULL and the consumer targets the same NULL bucket. Verified by code inspection of `test_builds.py`, `test_reservations.py`, `test_meta_parts.py`.
- [ ] Manual post-deploy smoke: receive an order, add stock, build consume on a project — common ledger flows still work.

No migration. No ops step. Independent of PRs #26 and #27.

### DIFF
--- a/backend/app/domain/builds/service.py
+++ b/backend/app/domain/builds/service.py
@@ -23,7 +23,7 @@ from app.domain.lots.models import Lot
 from app.domain.parts.models import Part, PartMetaMember, PartSubstitute
 from app.domain.projects.models import Project, ProjectEntry
 from app.domain.stock.models import StockEntry
-from app.domain.stock.service import current_quantity
+from app.domain.stock.service import current_quantity, lock_parts_for_stock_write
 from app.domain.storage.models import StorageLocation
 
 
@@ -162,9 +162,20 @@ def apply_reservations(
     only by (build_id, part_id, project_id, entry quantity) — no storage
     location or lot is bound, since the consumer picks those at consume-time.
     """
+    # BE2-008: lock every distinct part_id we're about to touch, in
+    # deterministic order, before any read. Without this, concurrent
+    # apply_reservations / release_reservations / consume on overlapping
+    # BOMs race on the reserved ledger and can either double-write or
+    # under-write the counter rows.
+    consumable = _consumable_entries(db, workspace_id=workspace_id, project=project)
+    lock_parts_for_stock_write(
+        db,
+        workspace_id=workspace_id,
+        part_ids=[e.part_id for e in consumable if e.part_id is not None],
+    )
     now = _now()
     written = 0
-    for e in _consumable_entries(db, workspace_id=workspace_id, project=project):
+    for e in consumable:
         part = db.get(Part, e.part_id)
         if part is None:
             continue
@@ -210,6 +221,15 @@ def release_reservations(
     )
     if not reserve_rows:
         return 0
+    # BE2-008: aggregate distinct part_ids from the reserve set, sort,
+    # take the per-part lock before reading the release-counter set.
+    # This serialises archive-while-consuming and any release flow that
+    # races with another build releasing the same parts.
+    lock_parts_for_stock_write(
+        db,
+        workspace_id=workspace_id,
+        part_ids=[r.part_id for r in reserve_rows],
+    )
     released_ids = set(
         db.execute(
             select(StockEntry.related_entry_id)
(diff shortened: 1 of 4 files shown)


matescb (commented):
## Approve with notes

**Closes:** `BE2-001`, `BE2-008`, `BE-002`, `DB-002`. Fills three gaps in the stock-write advisory-lock coverage and aligns service-layer validation with the `0013` non-negative trigger's NULL-bucket grouping.

### What the patch does
1. New helper `lock_parts_for_stock_write(db, workspace_id, part_ids)` — takes per-`(ws, part)` advisory locks in deterministic UUID-string sort order (deadlock-safe).
2. Adds the lock to `add_stock` (producer side — `PR #11` left this off).
3. Bundles bulk locks at `orders.receive`, `builds.apply_reservations`, `builds.release_reservations`, and `builds.consume` — each acquires the full superset of `part_id`s its transaction will touch.
4. Adds `bucket_match: bool = False` parameter to `current_quantity()`. When `True`, `lot_id=None` / `storage_location_id=None` mean "the SQL NULL bucket specifically" (`IS NULL`), aligning with the trigger's `IS NOT DISTINCT FROM` semantics. `remove_stock`, `move_stock`, `adjust_stock`, and `consume`'s pre-pass + per-line checks now pass `bucket_match=True`.

### Strengths
- ✅ **Re-entrancy is correct**: `pg_advisory_xact_lock` is reentrant within a transaction, so `consume`'s superset acquisition + nested `release_reservations`'s own acquisition is safe (the inner is a no-op for already-held part IDs).
- ✅ **Deterministic ordering** (`sorted(set(part_ids), key=str)`) prevents AB/BA deadlocks across concurrent transactions touching overlapping supersets. UUID `__str__` is canonical so cross-process ordering is stable.
- ✅ **`bucket_match` design**: opt-in flag preserves global-aggregate semantics for reports/list endpoints (`total_for_part` callers) while mutating paths get trigger-aligned per-bucket semantics. Default is the safer "don't break callers".
- ✅ **`build.consume` bundle covers BOM ∪ lines ∪ output_part** — the `associated_subassembly_part_id` write is locked, closing the producer-side hole on sub-assembly output.
- ✅ Justification comments are clear about *why* the producer side now needs the lock (invariant reads + trigger fall-through producing 500s instead of 4xx).

### Notes (non-blocking)

1. **No concurrency regression test added** — `TEST-003` from the v2 teardown still applies. The new `test_add_stock_takes_advisory_lock` self-describes as "smoke check that doesn't actually verify the lock". Recommendation for a follow-up: add a threaded test with a `threading.Barrier` for two concurrent `add_stock` calls on the same `(ws, part)` and assert both land cleanly (no trigger fall-through, no 500s). This is the regression net for the new producer-side lock and also exercises the deadlock-free ordering claim.
2. **`orders.receive` lock scope is intentionally broader than the touched lines**: covers every entry on the order even if the receive only addresses a subset. Comment justifies as "symmetric coverage" — fine, but a single-line receive on a 50-line order takes 50 locks. Performance is irrelevant for typical orders; keep an eye on the "import 5000-line order" path.
3. **`current_quantity(bucket_match=True)` boundary**: a future caller passing `bucket_match=True` with `lot_id=None` against a part whose stock is partly in `lot=None` and partly in `lot=L1` will see only the NULL-bucket sum — correct trigger view, but a one-line docstring caveat would prevent the assumption "no lot specified = total available".
4. **Service-layer tests bypass the API path** (`add_stock(db, ...)` directly with hand-built models). Existing concurrency tests use the HTTP path; the new tests are pure service-layer. Mixed approach is fine, but if `_belongs(part, workspace_id)` ever changes shape, the service-layer tests won't catch route-side regressions.

PR #29, opened 2026-05-02
feat(infra): pre-deploy pg_dump + post-deploy health gate; /api/health checks DB + uploads

matescb:
## Summary

Closes the deploy-safety gap that v1 Infra CRIT-1 / HIGH-1/7/8 and v2 `INFRA2-001` + `INFRA2-002` (both Critical) surface from the same root cause: auto-deploy on `main` ships a destructive migration in seconds with no pre-migration snapshot and no post-deploy verification.

Three coupled changes:

1. **`/api/health`** (`backend/app/main.py`) executes `SELECT 1` against the engine and checks `UPLOAD_DIR` is a writable directory. Returns 200 with `{db, uploads}` detail on success, 503 with the same shape on failure. Preserves the `{data, status}` envelope.
2. **`docker-compose.prod.yml`** — new backend healthcheck (curl /api/health, 60s start_period for alembic upgrade) and `web.depends_on.backend.condition: service_healthy`. Closes the post-deploy 502 window where nginx served before the backend was ready.
3. **CI deploy script** (`.github/workflows/ci.yml`) — pre-deploy `pg_dump` via new `deploy/predeploy-dump.sh` (fails loud, dump must land at `/srv/backups/stockmanager/pre-deploy-${TS}-${SHA}.sql.gz`); post-deploy 30 × 5s curl loop against the public health endpoint, with backend log tail on failure.

## Test plan

- [ ] CI green — new `backend/tests/test_health.py` covers 200 happy path + 503 on unwritable uploads + 503 on DB exception + envelope shape.
- [ ] `bash -n deploy/predeploy-dump.sh` — syntax verified locally.
- [ ] Manual post-deploy: dump file appears at `/srv/backups/stockmanager/pre-deploy-*.sql.gz`; `docker compose ps` shows backend as `healthy`; CI run logs show the curl loop succeeding within ~10–30s.
- [ ] Manual rollback rehearsal: pick a recent pre-deploy dump and verify it loads cleanly into a throwaway local Postgres (`gunzip -c <dump> | psql ...`).

## Ops considerations

- The dump runs against the **running** db container, BEFORE `docker compose up --build`. If the dump fails, the deploy aborts and the existing containers keep serving the old code — exactly the rollback path that didn't exist before.
- The healthcheck's `start_period: 60s` covers cold-boot alembic upgrade; the curl loop's 30 × 5s covers the rebuild + warm-up window.
- Dumps are pruned to the last 14 (separate retention from the nightly cron at `deploy/backup.sh`).
- No env var changes. No schema change. No ops step before merge — this just adds checks.
- Independent of #26 / #27 / #28.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -121,5 +121,39 @@ jobs:
             # deploys.
             export SENTRY_RELEASE
             SENTRY_RELEASE=$(git rev-parse --short=12 HEAD)
+
+            # Pre-deploy DB snapshot (INFRA2-001 / Infra CRIT-1).
+            # Runs before `docker compose up` so the pre-migration state
+            # of the running Postgres is captured. Fails loud — if the
+            # dump can't be written, the deploy aborts here and the old
+            # containers keep serving the old code.
+            bash deploy/predeploy-dump.sh "${SENTRY_RELEASE}"
+
             docker compose -f docker-compose.prod.yml --env-file .env.prod up -d --build
             docker image prune -f
+
+            # Post-deploy health gate (INFRA2-002 / Infra HIGH-1/7/8).
+            # Polls the public health endpoint until 200 — covers the
+            # rebuild window plus alembic upgrade plus first-request
+            # warm-up. The 30 × 5s = 150s ceiling is generous for our
+            # one-worker stack; tighten if it routinely takes less.
+            # On failure the SSH script exits non-zero, which fails the
+            # deploy job and surfaces in GitHub Actions instead of
+            # leaving prod broken with a green CI badge.
+            echo "post-deploy health gate"
+            ok=0
+            for i in $(seq 1 30); do
+              if curl -fsS https://parts.matescb.cz/api/health > /dev/null 2>&1; then
+                echo "  /api/health OK after ${i} attempts ($((i*5))s)"
+                ok=1
+                break
+              fi
+              sleep 5
+            done
+            if [ "${ok}" -ne 1 ]; then
+              echo "post-deploy health gate FAILED — last response:"
+              curl -v https://parts.matescb.cz/api/health || true
+              echo "--- backend logs (tail) ---"
+              docker compose -f docker-compose.prod.yml --env-file .env.prod logs --tail=80 backend || true
+              exit 1
+            fi
--- a/backend/app/main.py
+++ b/backend/app/main.py
@@ -171,4 +171,47 @@ app.include_router(catalog.router, prefix="/catalog", tags=["catalog"])
 
 @app.get("/api/health")
 def health():
-    return {"data": {"status": "ok"}, "status": {"category": "ok", "message": "OK"}}
+    """Liveness + DB + uploads-volume check.
+
+    Used by three callers:
+    - the docker-compose backend healthcheck (controls when `web` starts,
+      and what `docker compose ps` reports for status),
+    - the post-deploy CI gate (`curl /api/health` retried after `docker
+      compose up` so a failed migration / missing env / DB outage fails
+      the deploy instead of returning a green CI result on a broken prod),
+    - manual smoke after operator-driven changes.
+
+    Returns 503 with structured detail when a check fails so the caller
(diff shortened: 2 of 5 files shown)


matescb (commented):
## Approve with notes

**Closes:** `INFRA2-001` (pre-deploy DB dump), `INFRA2-002` (post-deploy health gate), `INFRA-001` (health endpoint upgraded).

### What the patch does
- `deploy/predeploy-dump.sh` runs from the SSH step before `docker compose up`, gzip-pipes `pg_dump` to `/srv/backups/stockmanager/pre-deploy-${TS}-${SHA}.sql.gz`, aborts deploy on dump failure or empty output, retains the last 14 dumps.
- `/api/health` now executes `SELECT 1` via `get_engine()` and `os.access(UPLOAD_DIR, os.W_OK)`; returns 503 with structured `{db, uploads}` detail on failure. Envelope contract preserved on both paths.
- `docker-compose.prod.yml` adds a backend `healthcheck` (curl-based, `start_period: 60s`) and gates `web` on `condition: service_healthy` — closes the post-deploy 502 window.
- CI deploy step polls `https://parts.matescb.cz/api/health` for 30 × 5s = 150s after `up -d`; dumps `docker compose logs --tail=80 backend` and exits non-zero on failure.
- Four new pytest cases pin the happy + 503 paths and the envelope shape.

### Strengths
- ✅ Each ID has a verifiable code change, not just a doc edit; the spec-to-implementation mapping in `INFRA2-001`/`002` is followed almost line-for-line.
- ✅ Health endpoint stays anonymous, doesn't leak version/env, isn't decorated by slowapi — all correct.
- ✅ `curl` is already installed in `backend/Dockerfile`, so the compose healthcheck won't `OCI exec failed`.
- ✅ The dump script's "is `db` actually running?" precheck and the `.tmp` → atomic `mv` rename are nice touches.
- ✅ Aborting deploy on dump failure is the right behaviour and is correctly enforced by `set -euo pipefail` plus the empty-file guard.

### Notes (non-blocking)
- ⚠️ `pg_dump` uses plain SQL + gzip. `--format=custom` (or `directory`) gives selective restore, parallel restore, and smaller artefacts at no real cost. Worth changing now while only 14 dumps exist.
- ⚠️ `test_health_returns_503_when_db_query_raises` patches `app.infra.db.get_engine`, but `main.health()` does `from app.infra.db import get_engine` *inside* the function, so the patch binds to the source module — that does work, but it's load-bearing on import-time-vs-call-time semantics. A quick `monkeypatch.setattr("app.infra.db.get_engine", ...)` is identical and clearer; the `patch("app.main.os.access", wraps=lambda ...)` line is also overly clever (`return_value=True` would do).
- ⚠️ `INFRA2-016` (encrypted backups) is correctly out of scope, but the new pre-deploy dumps land on the same VPS in plaintext SQL — same exposure as the nightlies. Worth a tracking note in `docs/deployment.md`.
- ⚠️ Spec called for an `nginx`/`web` healthcheck too (`INFRA2-002` step 2). This PR only adds the backend one. Acceptable since `web` is gated on backend health and the deploy gate hits the public URL through nginx, but flag it as a known follow-up.
- ⚠️ The post-deploy gate hits the public hostname over TLS — so it transitively also checks Apache + cert + DNS. Good. But `parts.matescb.cz` is hard-coded in the workflow; consider `${{ secrets.DEPLOY_HEALTH_URL }}` for parity with other deploy secrets.
- ⚠️ `docs/deployment.md` rollback section still says "restore from a `pg_dump` taken before the deploy"; it should now point at `/srv/backups/stockmanager/pre-deploy-*.sql.gz` and the 14-dump retention. (`INFRA2-001` fix instruction explicitly asked for this doc update.)

### Ops notes
First post-merge deploy will exercise the script for real. Watch for: (1) `docker compose exec -T db pg_dump` permissions if the local `pg_hba.conf` ever drifts off `trust`; (2) the 60s `start_period` is generous now but the alembic chain is growing — revisit if a future migration takes longer; (3) `/srv/backups/stockmanager` must be writable by the `deploy` user — already true for the nightly cron, so should be fine, but the first run is the test.

PR #30, opened 2026-05-02
feat(db): FKs + partial unique + pg_trgm search indexes (DB-001/003/004 + BE2-018)

matescb:
## Summary

Closes the four DB-layer items from the 2026-05 review:

| ID | What |
|---|---|
| `DB-001` / `BE2-002` / v1 `BE CRIT-4` | Add FKs (`ON DELETE SET NULL`) on `stock_entries.{order_id, order_entry_id, build_id}` and `lots.{source_order_id, source_build_id}`. They were bare UUIDs with no integrity. |
| `DB-003` | Replace full `UniqueConstraint` on `storage_locations` / `tags` `(workspace_id, name)` with partial unique indexes `WHERE archived_at IS NULL`, so archiving a row frees up its name (mirrors the `parts.mpn` pattern from alembic 0011). |
| `DB-004` | Add `(workspace_id, archived_at)` partial composite indexes on `attachments`, `tag_links`, `custom_fields`, `bom_import_presets`, `project_entries`. |
| `BE2-018` | Enable `pg_trgm` and add GIN trigram indexes on the 6 search-hot name/mpn columns (`parts.name`, `parts.mpn`, `storage_locations.name`, `projects.name`, `lots.name`, `orders.name`). |

Migration is `0018_db_schema_cleanup.py`, `revision="0018"`, `down_revision="0017"` (Batch 1).

## Migration safety

* **FK pre-flight**: orphan UUIDs in the cross-table-ref columns are `UPDATE … SET NULL` before the FK is created. The user reports no real prod data, but the pattern is the right one for live data and `tests/test_db_schema.py` exercises it.
* **Partial-unique pre-flight**: the migration `SELECT`s for active-row name duplicates on `storage_locations` / `tags` and **raises `RuntimeError` with a clear message** if any exist. Operator must reconcile (rename or archive) before retrying.
* **Idempotent extension**: `CREATE EXTENSION IF NOT EXISTS pg_trgm`. Downgrade does **not** drop the extension (harmless to leave installed; future migrations may rely on it).
* **Round-trip verified**: read the migration file end-to-end — every `op.create_*` in `upgrade()` has a matching `op.drop_*` in `downgrade()`, and the unique constraints are restored. `alembic upgrade 0018 && alembic downgrade 0017 && alembic upgrade 0018` is symmetric.

## Pre-deploy `pg_dump` checklist

This migration adds FKs and rewrites unique constraints — irreversible w.r.t. orphan-row data even with the pre-flight in place. **Before merging**:

- [ ] PR #29 (predeploy-dump-and-health-gate) merged → CI takes the dump automatically. **If #29 has not landed**, take a manual `pg_dump` per `docs/deployment.md#backups` *before* approving this PR for merge to `main`.
- [ ] Confirm with the operator that `RuntimeError` on active duplicate (workspace_id, name) rows is acceptable behaviour for prod (it is — failing loud > silently mid-run).

## v2 IDs closed

`DB-001`, `DB-003`, `DB-004`, `BE2-002`, `BE2-018`, plus the v1 carryover `BE CRIT-4`.

## Test plan

- [ ] CI runs `pytest backend/tests/test_db_schema.py` — covers FK existence + ondelete='SET NULL', partial-unique semantics on storage/tags, composite-index existence, pg_trgm extension installed, and trgm index defs.
- [ ] CI runs the full backend suite to ensure no regression on existing FK-dependent flows.
- [ ] Manual: against a fresh dev DB, `alembic upgrade 0018 && alembic downgrade 0017 && alembic upgrade 0018` round-trips cleanly.

### DIFF
--- a/backend/alembic/versions/0018_db_schema_cleanup.py
+++ b/backend/alembic/versions/0018_db_schema_cleanup.py
@@ -1,0 +1,249 @@
+"""DB schema cleanup: FKs on stock_entries cross-table refs, partial unique
+on soft-delete tables, composite (workspace_id, archived_at) indexes,
+pg_trgm GIN indexes for ILIKE search.
+
+Revision ID: 0018
+Revises: 0017
+Create Date: 2026-05-02
+
+Closes:
+  * DB-001 / BE2-002 / v1 BE CRIT-4 — `stock_entries.order_id`,
+    `order_entry_id`, `build_id` and `lots.source_order_id`,
+    `source_build_id` were bare UUID columns with no FK enforcement.
+    A cascading delete of an order/build left dangling pointers.
+  * DB-003 — `uq_storage_ws_name` and `uq_tag_ws_name` were full
+    UniqueConstraints, ignoring the soft-delete pattern. Restoring an
+    archived row clashed with re-using its name. Replaced with partial
+    unique indexes `WHERE archived_at IS NULL`.
+  * DB-004 — Several workspace-scoped tables (`attachments`,
+    `tag_links`, `custom_fields`, `bom_import_presets`,
+    `project_entries`) lacked the universal `(workspace_id, archived_at)`
+    index that every active-row query relies on.
+  * BE2-018 — Search ILIKE %q% on parts/storage/projects/lots/orders
+    was a 5x sequential scan per keystroke. pg_trgm GIN indexes turn
+    those into index-supported lookups.
+
+Pre-flight:
+  * Orphan UUIDs in the cross-table-ref columns are NULLed before the
+    FK is added. The user reports no real prod data, but the pattern
+    is correct and the tests exercise it.
+  * Active-row name duplicates on storage_locations / tags would block
+    the partial unique index. The migration RAISES with a clear error
+    message — the operator must reconcile data and retry.
+
+Idempotence / round-trip:
+  * `CREATE EXTENSION IF NOT EXISTS pg_trgm` is a no-op when present.
+  * Index creation uses `op.create_index` (errors if it already exists);
+    the migration is fresh, so no IF NOT EXISTS dance is needed.
+  * Downgrade restores prior FK-less columns + UniqueConstraints, but
+    deliberately does NOT drop the pg_trgm extension (harmless to leave
+    installed and other migrations may rely on it later).
+"""
+from __future__ import annotations
+
+from alembic import op
+import sqlalchemy as sa
+
+
+revision = "0018"
+down_revision = "0017"
+branch_labels = None
+depends_on = None
+
+
+# Tables that get a partial composite (workspace_id, archived_at) index
+# in this migration. The 5 tables identified by DB-004. All inherit
+# `archived_at` via the WorkspaceOwned mixin.
+_ARCHIVED_AT_INDEX_TABLES = (
+    "attachments",
+    "tag_links",
+    "custom_fields",
+    "bom_import_presets",
+    "project_entries",
+)
+
+
+# pg_trgm GIN indexes. Each tuple: (index_name, table, column).
+_TRGM_INDEXES = (
+    ("ix_parts_ws_name_trgm", "parts", "name"),
+    ("ix_parts_ws_mpn_trgm", "parts", "mpn"),
+    ("ix_storage_ws_name_trgm", "storage_locations", "name"),
+    ("ix_projects_ws_name_trgm", "projects", "name"),
+    ("ix_lots_ws_name_trgm", "lots", "name"),
+    ("ix_orders_ws_name_trgm", "orders", "name"),
+)
+
+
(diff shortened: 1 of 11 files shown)


matescb (commented):
## Approve with notes

**Closes:** `DB-001` (FKs on cross-table refs), `DB-003` (partial unique on soft-delete tables), `DB-004` (composite `(workspace_id, archived_at)` indexes), `BE2-018` (pg_trgm GIN indexes for ILIKE search) — index-side only.

### What the patch does
One migration (`backend/alembic/versions/0018_db_schema_cleanup.py`) plus eight model files mirror the schema. It adds 5 `ON DELETE SET NULL` FKs (`stock_entries.{order_id, order_entry_id, build_id}`, `lots.{source_order_id, source_build_id}`), converts `uq_storage_ws_name` and `uq_tag_ws_name` to partial unique indexes `WHERE archived_at IS NULL`, adds 5 `(workspace_id, archived_at)` partial composites, and creates the `pg_trgm` extension + 6 GIN trigram indexes. `backend/tests/test_db_schema.py` (247 lines) pins all of it against `pg_catalog`.

### Strengths
- ✅ Pre-flight orphan-NULLing for the FKs and an explicit `RuntimeError` if active duplicate `(workspace_id, name)` rows exist before the partial unique is created (`0018_db_schema_cleanup.py:128-145`) — fail-loud is the right call.
- ✅ Constraint names are pinned on both the model (`name="fk_stock_entries_order_id"`, ...) and the migration, so `downgrade()` can drop by name.
- ✅ `uq_workspace_member` is correctly skipped — its dedup is by `status` not `archived_at` (matches `docs/teardown/database.md:134`).
- ✅ `uq_storage_ws_name` / `uq_tag_ws_name` keep the same constraint name when re-created as an index, which is the same pattern as `uq_parts_ws_mpn` (the canonical template named in CLAUDE.md).
- ✅ Tests run against a real Postgres via `alembic upgrade head` (`backend/tests/conftest.py:38`), so model-vs-migration drift surfaces in CI. Backend-tests are green on this branch.
- ✅ Downgrade restores all 5 FKs, both UniqueConstraints, and drops the trigram indexes; pg_trgm is intentionally left installed (documented in the docstring).

### Notes (non-blocking)
- ⚠️ `BE2-018` is only **half-fixed**: the GIN indexes land but `backend/app/api/routes/search.py` is unchanged. `gin_trgm_ops` does accelerate `ILIKE '%q%'` (that's the point of pg_trgm), so the index will be used — but the rest of the BE2-018 fix instructions are still pending: `q` has no `max_length=200`, no per-IP rate limit, no outer total cap. Worth a follow-up issue or a sentence in the PR description so it doesn't get marked closed.
- ⚠️ All `op.create_foreign_key` calls in section 1 take an `AccessExclusive` lock + full table scan to validate. Fine at current data volume; at any real scale the safer pattern is `ADD CONSTRAINT ... NOT VALID` followed by `VALIDATE CONSTRAINT` in a separate migration. Not blocking — flag for the runbook.
- ⚠️ `op.create_index` (sections 3+4) runs in-transaction and takes a `ShareLock` on each table. Six GIN indexes on `parts`/`orders`/`lots`/etc. will block writes for the build duration. Standard prod-safe alternative is `CREATE INDEX CONCURRENTLY` inside `with op.get_context().autocommit_block(): ... postgresql_concurrently=True`. Same call: fine at current volume, worth knowing.
- ⚠️ Composite list (`_ARCHIVED_AT_INDEX_TABLES`) deliberately omits `stock_entries` — correct, it doesn't inherit `WorkspaceOwned` and has no `archived_at`, despite the teardown listing it. A one-line code comment to that effect would save the next reader the lookup.
- ⚠️ `lots/models.py` import reorder is cosmetic; harmless but unrelated.

### Migration ordering / rebase note
**The user's note is still pending and accurate.** `0018_db_schema_cleanup.py:62` has `down_revision = "0016"`. PR #31 (`fix/auth-csrf-hardening`) adds `backend/alembic/versions/0017_session_token_hash.py`. Whichever PR merges second must rebase: if #31 lands first (likely, given the numbering), this PR's `down_revision` must change from `"0016"` to `"0017"` before merge — otherwise alembic will fork on `0016`. The migration's own docstring (lines 12–17) already calls this out, which is great. **Don't merge before rebasing on top of post-#31 main.**

### Ops notes
- Deploy is auto on merge to `main` and runs `alembic upgrade head` on container start. Pre-flight `pg_dump` per `docs/deployment.md#backups` is sensible given five new FKs and two constraint swaps. The pre-flight-duplicates check raising mid-migration would leave the FKs already added but the partial-unique work uncommitted — alembic's per-revision transaction should roll the FKs back, but worth eyeballing the prod logs on deploy.
- pg_trgm is in `postgres-contrib`, ships with the official `postgres:` image used in compose — no infra change needed.

PR #31, opened 2026-05-02
fix(security): CSRF middleware + session token hash + workspace switch validation (SEC2-001/003/004/012/015 + BE2-004)

matescb:
## Summary

Batch 1 of the 2026-05-01 v2 review remediation. Closes the auth and
cookie-auth-boundary gaps that were stacking up since the original
2026-04-30 review.

### v2 IDs closed

| ID | Severity | What |
|---|---|---|
| `SEC2-001` | Critical | CSRF Origin/Referer middleware on every cookie-authenticated state-changing endpoint |
| `SEC2-003` | High | Session tokens are now hashed at rest (mirror of PR #14 invitations fix) |
| `SEC2-004` | High | `/api/workspaces/{id}/switch` validates UUID + requires `CurrentUser` + checks active membership |
| `SEC2-012` / `BE2-017` | Medium | Scanner-license-key endpoint now requires `member+` (viewer is blocked) |
| `SEC2-015` | Medium | Session rotation on login + sliding 24h idle expiry |
| `BE2-004` | High | Authenticated user can no longer mint unbounded workspaces (cap 5, 10/hour rate limit) |
| v1 Sec CRIT-4 | Critical (escalated) | Workspace-switch cookie tightened from `SameSite=Lax` to `SameSite=Strict` |

### Migration

`0017_session_token_hash.py` (`revision="0017"`, `down_revision="0016"`).

**Pre-merge note: all existing sessions are invalidated by this
migration.** The schema swaps `user_sessions.token` (plaintext PK)
for `user_sessions.token_hash` (SHA-256 PK), and there is no
migration path that keeps current sessions valid without storing
plaintext on the server — which is the whole point of the change.
Every active user will be force-logged-out at the next request and
must re-login. This is acceptable: prod is dev-grade and there is
no real production data yet.

### Scoped-down / deferred items

- The plan asked for a blanket `dependencies=_member_gate` on the
  workspaces router. I did **not** apply it — `require_member_for_writes`
  blocks viewers from any write to the *current* workspace, which
  would also block legitimate flows (a viewer in workspace A still
  needs to be able to create / switch into their own workspace B).
  The named target (viewer reaching `/scanner-license-key`) is closed
  by the explicit per-endpoint `Depends(require_role("member"))`.
- The plan suggested dropping the `try/except ValueError` UUID parse
  in `core/deps.py`. I kept it: the switch route now parses UUID
  upstream, but the `X-Workspace-Id` header is still untrusted client
  input and a malformed value must produce a clean 4xx (via fall-through
  to the first membership), not a 500. Comment updated to reflect why.

### Test plan
- [ ] CI: `pytest` (no docker locally) — covers test_session_hashing,
      test_csrf, test_workspaces, test_workspace_quota, the new
      viewer-license-key case in test_rbac_viewer, and the updated
      `samesite=strict` assertion in test_security_hardening.
- [ ] CI: alembic migration round-trip (`upgrade 0017 → downgrade 0016
      → upgrade 0017`) is exercised by `tests/conftest.py::engine`
      starting every session at the migration head; the downgrade
      path is structurally reversible (table is wiped on the way
      down because plaintext tokens cannot be recovered).
- [ ] Post-merge: confirm prod re-login works after auto-deploy
      applies 0017 (session rotation + new cookie format).

### DIFF
--- a/backend/alembic/versions/0017_session_token_hash.py
+++ b/backend/alembic/versions/0017_session_token_hash.py
@@ -1,0 +1,107 @@
+"""session tokens stored hashed at rest + sliding expiry
+
+Revision ID: 0017
+Revises: 0016
+Create Date: 2026-05-02
+
+Mirror of the invitation token hashing landed in 0014, applied to
+`user_sessions` (SEC2-003). The plaintext session token used to be
+the primary key of every row, so a DB dump leaked every active
+session as a replayable cookie. We now store only `sha256(token)` in
+`token_hash`. The plaintext lives only on the client cookie and is
+never persisted.
+
+Plus a new `last_used_at` column for sliding expiry (SEC2-015): the
+deps layer bumps it on every successful auth lookup, and rejects
+sessions idle for longer than the configured window (24h at the
+time of writing) even when the absolute `expires_at` is still in
+the future.
+
+Behaviour:
+- Drop the `user_sessions.token` column entirely. We don't try to
+  backfill `token_hash` from existing rows because the schema swap
+  is destructive (the old PK goes away) and the user pool is small
+  enough that forcing every active session to re-login is acceptable.
+  All existing sessions are invalidated by this migration — that's
+  the documented trade-off.
+- Add `token_hash CHAR(64) PRIMARY KEY`.
+- Add `last_used_at TIMESTAMPTZ NOT NULL DEFAULT now()`.
+
+Downgrade is structurally reversible (re-add `token` column + drop
+`token_hash` / `last_used_at`) but the original plaintext tokens are
+unrecoverable. Rolling back further than 0016 forces every session
+to re-login a second time — fine, since 0017 already invalidated
+them all.
+"""
+from __future__ import annotations
+
+from alembic import op
+import sqlalchemy as sa
+
+
+revision = "0017"
+down_revision = "0016"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # Wipe every existing row so the swap from `token`-keyed to
+    # `token_hash`-keyed is unambiguous. There is no migration path
+    # that keeps current sessions valid without storing plaintext on
+    # the server, and that's the whole point of this change. All
+    # users will be forced to re-login post-deploy — pre-merge note
+    # in the PR body warns operators.
+    op.execute("DELETE FROM user_sessions")
+
+    # Drop the old plaintext PK column. PK constraint name is
+    # `user_sessions_pkey` by Postgres default, dropped by `op.drop_column`
+    # automatically.
+    op.drop_column("user_sessions", "token")
+
+    # Add the new hashed PK. CHAR(64) is the exact length of a hex
+    # SHA-256 digest; using VARCHAR(64) keeps schema-comparison tools
+    # quieter without changing storage in Postgres.
+    op.add_column(
+        "user_sessions",
+        sa.Column("token_hash", sa.String(length=64), nullable=False),
+    )
+    op.create_primary_key(
+        "user_sessions_pkey",
+        "user_sessions",
+        ["token_hash"],
+    )
+
+    # Sliding-expiry tracking. `server_default=text('now()')` so the
+    # column is populated on every existing row at upgrade time —
+    # though there are no rows after the DELETE above, this keeps the
+    # migration idempotent for any test/dev DB that re-applies it.
+    op.add_column(
+        "user_sessions",
+        sa.Column(
+            "last_used_at",
+            sa.DateTime(timezone=True),
+            nullable=False,
+            server_default=sa.text("now()"),
+        ),
+    )
+
+
+def downgrade() -> None:
+    """Structurally reversible. The original plaintext tokens are gone
+    forever; downgraded sessions cannot be authenticated until the
+    user logs in again. Wipe the table on the way down so the
+    re-introduced `token NOT NULL` PK is clean."""
+    op.execute("DELETE FROM user_sessions")
+    op.drop_column("user_sessions", "last_used_at")
+    op.drop_constraint("user_sessions_pkey", "user_sessions", type_="primary")
+    op.drop_column("user_sessions", "token_hash")
+    op.add_column(
+        "user_sessions",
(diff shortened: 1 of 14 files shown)


matescb (commented):
## Approve with notes

**Closes:** `SEC2-001`, `SEC2-003`, `SEC2-004`, `SEC2-012` / `BE2-017`, `SEC2-015`, `BE2-004`, plus v1 `Sec CRIT-4` (workspace-switch cookie tightened to `SameSite=Strict`).

### What the patch does
- **CSRF** (`backend/app/main.py:152-220`): `CsrfOriginMiddleware` rejects POST/PATCH/PUT/DELETE whose `Origin` (or `Referer` fallback) doesn't normalise into `cors_origin_list`. Exempts `GET/HEAD/OPTIONS`, `/api/auth/login`, `/api/auth/signup`, `/api/sentry-tunnel`. No double-submit token; relies on browsers being trustworthy senders of `Origin` — fine for an SPA, no FE changes needed.
- **Session hash at rest** (`backend/app/core/auth.py:71-95`, `backend/app/domain/users/models.py:32-44`, `backend/alembic/versions/0017_session_token_hash.py`): plaintext PK `token` is dropped and replaced with `token_hash CHAR(64)` = `sha256(cookie_token)`. Migration `DELETE`s every row first — invalidates all sessions on deploy (intended).
- **Workspace switch** (`backend/app/api/routes/workspaces.py:274-329`): typed `workspace_id: UUID` (so `non-uuid` → 422), requires `CurrentUser`, 404s unless `WorkspaceMember.status == "active"`, sets cookie with `SameSite=Strict`.
- **License key RBAC** (`backend/app/api/routes/workspaces.py:121-124`): `dependencies=[Depends(require_role("member"))]` on `GET /workspaces/current/scanner-license-key`.
- **Session rotation** (`backend/app/api/routes/auth.py:104-110`, `backend/app/core/deps.py:48-66`): `revoke_all_user_sessions(db, user.id)` on login; idle-rejection at 24h via `last_used_at` bumped in `get_current_user`.
- **Workspace creation cap** (`backend/app/api/routes/workspaces.py:51-81`): per-user `_OWNED_ORG_WORKSPACE_CAP = 5` on `kind="organization"` rows owned by caller (personal excluded), plus `@limiter.limit("10/hour")`. Returns 409 with `{existing_count, cap}`.

### Strengths
- ✅ Test coverage is thorough and goal-oriented: `tests/test_csrf.py` (7 cases incl. Referer fallback + login exemption), `tests/test_session_hashing.py` (4 cases incl. forced idle expiry), `tests/test_workspaces.py` (5 cases incl. non-member 404 doesn't leak existence), `tests/test_workspace_quota.py`, plus the `test_security_hardening.py` strict-cookie update.
- ✅ Clean test-harness retrofit: `backend/tests/conftest.py:20-46` injects `http://testserver` into `CORS_ORIGINS` and monkey-patches `TestClient.__init__` to send `Origin` by default — existing tests don't have to change. Nice.
- ✅ Hash digest as the PK lets the lookup be a single PK scan with no constant-time-compare gymnastics; rationale is explicitly documented (`core/auth.py:90-96`, `core/deps.py:35-37`).
- ✅ Strict cookie is correctly scoped to *workspace-switch* only; the session cookie stays Lax so login-style top-level redirects still work — explained in the route docstring.

### Notes (non-blocking)
- ⚠️ **Login is CSRF-exempt and session rotation is on the same path.** A foreign-origin POST to `/api/auth/login` succeeds today (`test_login_exempt_from_csrf` enshrines this). The classic risk is **session fixation / login-CSRF** — attacker logs the victim into the attacker's account. Brute-force is rate-limited and the new `revoke_all_user_sessions` actually mitigates the *other* direction nicely, but the login-CSRF surface itself isn't closed. Worth a follow-up: either keep an Origin check on `/login` too, or accept-and-document. Same applies to `/signup`.
- ⚠️ **No rotation on password change.** Spec for SEC2-015 mentions privilege/credential changes; there's no password-change endpoint in the codebase today, so this is moot — but if/when one lands, remember to call `revoke_all_user_sessions` (and ideally don't revoke the *current* session so the user stays logged in).
- ⚠️ **HMAC vs raw SHA-256.** `hashlib.sha256(token).hexdigest()` is fine here — the cookie is 48 bytes from `secrets.token_urlsafe`, well above the rainbow-table threshold. HMAC keyed on `SESSION_SECRET` would defend against a future regression where token entropy drops; cheap to add later.
- ⚠️ **`get_current_user` commits `last_used_at` on every authenticated request** (`core/deps.py:62-65`). One extra UPDATE per request and the bump survives even if the request handler then 500s. Both deliberate per the comment, but: (a) on a chatty polling endpoint this is throughput overhead worth measuring, (b) consider rounding to the minute / only writing if `now - last_used_at > 60s` to coalesce.
- ⚠️ **Workspace cap query** (`workspaces.py:60-69`) joins `WorkspaceMember` against `Workspace` filtered on `owner_user_id == user.id` AND `WorkspaceMember.user_id == user.id`. The membership join is redundant for ownership counting (owner is always a member); a bare `Workspace.query` with the same filters would be one row narrower. Cosmetic.
- ⚠️ **Migration downgrade** wipes the table again. Operationally fine, but if anyone runs `0017 -> 0016 -> 0017` the second downgrade silently logs nobody out (no rows to delete) — surprising-but-harmless.
- ⚠️ Header capture in `test_csrf.py` uses `headers={"Origin": ""}` to clear the auto-injected one; verify `httpx`/Starlette actually forwards the empty header rather than dropping it (test behavior asserts 403 so clearly works today, but a future httpx upgrade could change this).

### Ops notes
- 🚨 **All sessions invalidated on deploy**: every active user gets a 401 on their next request and must re-login. PR body acknowledges this; surfacing here so it doesn't get lost — coordinate with users (or at least be ready for the support pings) when this hits prod.
- The `last_used_at server_default=now()` backfill is a no-op because the upgrade `DELETE`s the table first; harmless on dev but means there is no "remember me" carry-over from pre-0017.
- Migration is destructive but reversible structurally — just not semantically. Take a `pg_dump` per `docs/deployment.md#backups` before merge as usual.

PR #32, opened 2026-05-02
fix(security): provider asset hardening + security headers + BOM caps + route gaps

matescb:
## Summary

Security-hardening + small-route-correctness batch (Batch 3 of v2 review remediation).

### Closes

- **SEC2-006** (High) — Provider asset download follows redirects, no host allow-list, permits SVG.
- **SEC2-011** (Medium) — `GET /api/parts/assets/...` lacks `nosniff`, serves user-influenced inline.
- **SEC2-007 + BE2-006** (High) — BOM import unbounded base64 -> memory exhaustion.
- **SEC2-009** (Medium) — Public catalog HTML rendered without CSP / nosniff / frame-options.
- **SEC2-010** (Medium) — SPA / API responses ship no security headers.
- **INFRA2-007** (High) — Apache vhost + TLS hardening not in repo.
- **BE2-013** (Medium) — `OrderEntryIn.quantity_ordered` accepts 0 / negative.
- **BE2-014** (Medium) — `archive_storage` doesn't refuse when storage holds positive on-hand stock.
- **v1 Sec HIGH-4** (High) — `/catalog/{token}` no rate-limit.

### Key changes

- `backend/app/domain/parts/services/assets.py` — explicit Mouser/DigiKey host allow-list + DNS-must-be-public-IP guard (`ipaddress.ip_address(...).is_global`); `httpx.Client(follow_redirects=False)`; `image/svg+xml` dropped from `_EXT_BY_MIME` and any SVG content-type or `.svg` URL suffix lands as `.bin`.
- `backend/app/api/routes/parts.py` (asset-serve route only) — `X-Content-Type-Options: nosniff` on every response; explicit `media_type` from a small ext->mime map; non-image MIMEs (PDF, opaque `.bin`) forced to `Content-Disposition: attachment` mirroring `attachments.py:209-214`.
- `backend/app/domain/projects/schemas.py` + `bom_import.py` — `Field(..., max_length=5_000_000)` on both BOM `text_b64` schemas; post-decode 4 MB ceiling -> 413; 10 000-row cap -> 422.
- `deploy/nginx-web.conf` — CSP (with `https://*.ingest.sentry.io` in `connect-src` for the Sentry tunnel), `X-Content-Type-Options`, `X-Frame-Options DENY`, `Referrer-Policy same-origin`, HSTS, `Permissions-Policy ()` — all `always`.
- `deploy/parts.matescb.cz-le-ssl.conf` (new) — :443 Apache vhost with Mozilla-intermediate cipher suite, HSTS preload, header set. Operator step documented at the top of the file.
- `backend/app/api/routes/catalog.py` — `@limiter.limit("60/minute")` on `catalog_html` + `catalog_json`; both responses carry the same security headers.
- `backend/app/domain/orders/schemas.py` — `OrderEntryIn.quantity_ordered: int = Field(..., ge=1)`; `OrderEntryPatch.quantity_ordered: int | None = Field(default=None, ge=1)`.
- `backend/app/api/routes/storage.py::archive_storage` — pre-flight `stock_for_storage`; returns 409 with `{message, blocking: [{part_id, lot_id, quantity}]}` when stock is still in residence.

### Tests added

- `backend/tests/test_provider_assets.py` (new) — host allow-list, private-IP / loopback / metadata refusal, redirect-refusal, SVG-lands-as-bin, served-image carries nosniff and stays inline, served-PDF / unknown-ext forces attachment.
- `backend/tests/test_archive_storage.py` (new) — empty storage archives, on-hand stock yields 409 with `blocking` detail, post-move-out archive succeeds.
- `backend/tests/test_bom_import.py` — schema cap, decoded-size cap (413), row-count cap (422).
- `backend/tests/test_orders.py` — quantity 0, negative, and PATCH-to-zero all 422.
- `backend/tests/test_catalog.py` — HTML and JSON responses carry the security headers.

## Operator step (required after merge)

Apache picks up the new file once the operator runs:

```
sudo cp deploy/parts.matescb.cz-le-ssl.conf \
        /etc/apache2/sites-available/parts.matescb.cz-le-ssl.conf
sudo apache2ctl configtest && sudo systemctl reload apache2
```

(See the top-of-file comment block for the rationale; certbot's renewal hook does **not** rewrite this file once it's been replaced.)

## Test plan

- [ ] `docker compose exec backend pytest backend/tests/test_provider_assets.py backend/tests/test_archive_storage.py backend/tests/test_bom_import.py backend/tests/test_orders.py backend/tests/test_catalog.py`
- [ ] Smoke: open a PDF datasheet via `/api/parts/assets/...?name=X` — saves as attachment, response carries `X-Content-Type-Options: nosniff`.
- [ ] Smoke: hit `/catalog/{token}` and check response headers include CSP + `X-Frame-Options: DENY`.
- [ ] Operator: `cp` the Apache file post-merge as above, run `apache2ctl configtest`, reload.

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

### DIFF
--- a/backend/app/api/routes/catalog.py
+++ b/backend/app/api/routes/catalog.py
@@ -7,16 +7,35 @@ from __future__ import annotations
 
 from html import escape
 
-from fastapi import APIRouter, HTTPException, status
-from fastapi.responses import HTMLResponse
+from fastapi import APIRouter, HTTPException, Request, status
+from fastapi.responses import HTMLResponse, JSONResponse
 from sqlalchemy import select
 
 from app.core.deps import DbSession
+from app.core.ratelimit import limiter
 from app.core.responses import ok
 from app.domain.parts.models import Part
 from app.domain.workspaces.models import Workspace
 
 router = APIRouter()
+
+
+# Same-origin headers for the public catalog responses (SEC2-009). The
+# nginx layer already attaches these for the SPA bundle, but the catalog
+# is fetched directly from FastAPI in some flows (e.g. the JSON endpoint
+# proxied through Apache without going through nginx) so we attach them
+# here too. Cheap and idempotent; nginx duplicates collapse.
+_CATALOG_HEADERS: dict[str, str] = {
+    "X-Content-Type-Options": "nosniff",
+    "X-Frame-Options": "DENY",
+    "Referrer-Policy": "same-origin",
+    "Content-Security-Policy": (
+        "default-src 'self'; img-src 'self' data:; "
+        "style-src 'self' 'unsafe-inline'; script-src 'none'; "
+        "frame-ancestors 'none'"
+    ),
+    "Permissions-Policy": "()",
+}
 
 
 def _resolve_workspace(db, token: str) -> Workspace:
@@ -134,19 +153,26 @@ def _render_html(ws: Workspace, parts: list[Part]) -> str:
 
 
 @router.get("/{token}", response_class=HTMLResponse)
-def catalog_html(token: str, db: DbSession):
+@limiter.limit("60/minute")
+def catalog_html(request: Request, token: str, db: DbSession):
     ws = _resolve_workspace(db, token)
     parts = _published_parts(db, ws.id)
-    return HTMLResponse(content=_render_html(ws, parts), status_code=200)
+    return HTMLResponse(
+        content=_render_html(ws, parts),
+        status_code=200,
+        headers=_CATALOG_HEADERS,
+    )
 
 
 @router.get("/{token}/parts.json")
-def catalog_json(token: str, db: DbSession):
+@limiter.limit("60/minute")
+def catalog_json(request: Request, token: str, db: DbSession):
     ws = _resolve_workspace(db, token)
     parts = _published_parts(db, ws.id)
-    return ok(
+    body = ok(
         {
             "workspace": {"id": str(ws.id), "name": ws.name},
             "parts": [_part_dict(p) for p in parts],
         }
     )
+    return JSONResponse(content=body, headers=_CATALOG_HEADERS)
--- a/backend/app/api/routes/parts.py
+++ b/backend/app/api/routes/parts.py
@@ -153,6 +153,22 @@ def _serialize(
 # ---------------------------------------------------------------------------
 
 
+# MIME-by-extension map for the serve route. Anything not in this set
+# is treated as an opaque binary and forced to download as an attachment
+# — which keeps a future provider-asset-MIME drift (e.g. an HTML page
+# erroneously saved with a .bin extension) from rendering inline. SVG is
+# intentionally absent; SEC2-006 / SEC2-011.
+_ASSET_MIME_BY_EXT: dict[str, str] = {
+    "jpg": "image/jpeg",
+    "jpeg": "image/jpeg",
+    "png": "image/png",
+    "gif": "image/gif",
+    "webp": "image/webp",
+    "pdf": "application/pdf",
+}
+_INLINE_EXTS: frozenset[str] = frozenset({"jpg", "jpeg", "png", "gif", "webp"})
+
+
 @router.get("/assets/{ws_id}/{filename}")
 def get_provider_asset(
     ws_id: UUID,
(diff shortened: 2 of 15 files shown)


matescb (commented):
## Approve with notes

**Closes:** `SEC2-006` (provider asset host allow-list + SSRF), `SEC2-007` / `BE2-006` (BOM payload caps), `SEC2-009` / `SEC2-010` (security headers, nginx + Apache), `SEC2-011` (asset serve nosniff + attachment), `BE2-013` (order qty >= 1), `BE2-014` (block archive of non-empty storage), `INFRA2-007` (canonical Apache vhost in repo).

### What the patch does
- **Provider asset hardening** (`backend/app/domain/parts/services/assets.py`): adds an explicit Mouser/DigiKey host allow-list, a `socket.gethostbyname` + `ipaddress.is_global` check that blocks RFC1918 / loopback / link-local / 169.254 metadata, flips `httpx.Client(follow_redirects=False)` so a 30x is treated as refusal, and routes both `image/svg+xml` Content-Type and `.svg` URL suffix into a `.bin` extension.
- **Asset serve route** (`backend/app/api/routes/parts.py:156-225`): introduces an explicit `_ASSET_MIME_BY_EXT` map; sets `X-Content-Type-Options: nosniff` on every response, an explicit `media_type` (no `application/octet-stream` MIME-confusion path), and forces `Content-Disposition: attachment` for anything not in `_INLINE_EXTS` (PDFs, opaque binaries). Header is set explicitly rather than via Starlette's `content_disposition_type` kwarg — good call (older Starlettes silently no-op).
- **Security headers**: nginx (`deploy/nginx-web.conf`) gets CSP / HSTS / XFO / nosniff / Referrer-Policy / Permissions-Policy on the SPA layer with `always` so they ride 4xx/5xx; new canonical Apache vhost `deploy/parts.matescb.cz-le-ssl.conf` sets HSTS preload (`max-age=63072000; includeSubDomains; preload`), TLS 1.2+1.3 only with the Mozilla intermediate cipher suite, and the same response-headers set. Catalog routes (`backend/app/api/routes/catalog.py:23-37`) attach the same headers in-app since they're sometimes hit through Apache without traversing nginx — sensible defence in depth.
- **BOM cap**: `BomImportPreviewIn.text_b64` and `BomImportCommitIn.text_b64` get `Field(max_length=6_000_000)`; `bom_import._decode_b64` raises 413 past 4MB decoded; `_enforce_row_cap` raises 422 past 10k rows. Layered intentionally — a 5 MB Field cap would foreclose the runtime check, the 6 MB / 4 MB pairing is documented in `schemas.py:70-77`.
- **Route gaps**: orders `quantity_ordered` enforced `>= 1` on POST and PATCH (`backend/app/domain/orders/schemas.py:16-30`); `archive_storage` returns structured 409 with `blocking: [{part_id, lot_id, quantity}]` when stock remains (`backend/app/api/routes/storage.py:113-134`); catalog HTML/JSON gain a `60/minute` rate limit.
- **Tests**: 13 new tests across `test_provider_assets.py`, `test_archive_storage.py`, `test_catalog.py`, `test_bom_import.py`, `test_orders.py`. Pre-existing `test_assets.py` correctly monkeypatches `_host_is_allowed` to keep the host check from intruding on unrelated content-addressing tests, with a comment explaining why.

### Strengths
- ✅ The host allow-list is a frozenset of explicit hostnames, not a substring/regex match — no `*.mouser.com` foot-gun that would let `mouser.com.evil.example` through.
- ✅ The SSRF check is the right shape: post-DNS `is_global` covers RFC1918, loopback, link-local AND the 169.254/16 AWS metadata range in one predicate. Tests cover all three.
- ✅ SVG defence is two-layered (Content-Type *and* URL suffix) and the file lands as `.bin`, which the serve route then forces to `attachment` — a malicious upstream cannot get inline rendering through any path.
- ✅ Layered BOM caps (Pydantic / decoded bytes / row count) with status codes deliberately split: 422 on schema, 413 on decoded size, 422 on rows.
- ✅ New Apache vhost has prominent `OPERATOR STEPS` block at the top — exactly the right place for this since `certbot --apache` regenerates over it.

### Notes (non-blocking)
- ⚠️ `Permissions-Policy: ()` (nginx line 32, Apache line 53) is **not a valid directive** — it has no feature name. It will be ignored by browsers, so it's a no-op rather than a hazard, but it gives a false sense of coverage. The scanner needs `camera=(self)`; recommend `Permissions-Policy: camera=(self), microphone=(), geolocation=(), payment=(), usb=()`. Flag this so a future "tighten the policy" pass doesn't accidentally break `web/src/components/scanner/ZxingScanner.tsx:176` (`navigator.mediaDevices.getUserMedia`) and the Scandit camera path.
- ⚠️ CSP at nginx line 33 omits `connect-src` for the Sentry tunnel itself (only `https://*.ingest.sentry.io`). The tunnel is same-origin (`/api/sentry-tunnel`, see `web/src/instrument.ts:34`) so it's covered by `default-src 'self'`, but the comment above the line suggests `*.ingest.sentry.io` is the live path — in practice it's the fallback. Worth tightening the comment or the directive once you decide which is real.
- ⚠️ `style-src 'self' 'unsafe-inline'` is acknowledged in the comment as a Vite-critical-CSS concession. Fine for now; the nonce migration is the right follow-up.
- ⚠️ The host allow-list is in code (`_ALLOWED_HOSTS` in `assets.py:55-69`), but the secrets/provider config is per-workspace. If you ever ship a third provider, you'll need to remember to grow this set as well as the catalog-keys split documented in `CLAUDE.md`. Maybe worth a brief comment cross-referencing the provider docs.
- ⚠️ `_host_is_allowed` issues a synchronous `socket.gethostbyname` on every fetch. httpx will resolve again for the actual GET — minor TOCTOU window where DNS could rebind between check and fetch. In practice `httpx` uses the same resolver and the upstream connection still has to come from a public IP, but a strictly correct fix is to resolve once, validate, then connect to the IP with a `Host:` header override. Not a blocker for the threat model here.
- ⚠️ Order `quantity_ordered >= 1` is the right call for new entries; double-check there are no migrations / fixtures with legacy `0` rows that a future PATCH would now fail to roundtrip. None spotted in this diff.

### Ops notes
- 📌 **Apache vhost re-install required on the VPS post-merge** (per author): `sudo cp deploy/parts.matescb.cz-le-ssl.conf /etc/apache2/sites-available/parts.matescb.cz-le-ssl.conf && sudo apache2ctl configtest && sudo systemctl reload apache2`. The auto-deploy only restarts the docker stack — Apache lives outside it. Worth adding a one-liner to `docs/deployment.md` next to the rollout notes so the next operator doesn't have to grep this PR description.
- HSTS preload is set with `preload` directive — once a browser sees this it's pinned for two years. Fine for `parts.matescb.cz` since it's already HTTPS-only behind certbot, but be aware: any subdomain that ever needs HTTP for any reason is now off the table.
- Browser CSP regressions to watch for: any `style="..."` attribute is allowed (covered by `'unsafe-inline'`), but any inline `<script>` would now be blocked. The SPA bundle is fully external (`<script type="module" src="/src/main.tsx">` only), so the migration should be transparent — but worth a manual smoke of the published catalog HTML page since `_render_html` in `catalog.py` builds an HTML doc and `script-src 'none'` there would block any future inline JS additions.
- Pre-existing tests bypass the host allow-list via monkeypatch — comment in `test_assets.py:34-37` explains why. Good hygiene; just be aware that the only real-network path is exercised by the new `test_provider_assets.py`, not the older suite.

CI green. Ship it after the `Permissions-Policy` fix above (or accept the no-op and document).

PR #33, opened 2026-05-02
fix(ui): global 401 + workspace-keyed cache + DataTable cleanup + scanner tokens

matescb:
## Summary

- Wires a global 401 handler at `QueryClient` construction so every authed query/mutation that 401s clears session state and bounces the user to `/login` with the original location preserved as `state.from`. Login + Signup replay that target on success.
- Workspace switch now flushes the React Query cache (`qc.clear()`) and navigates to `/parts` instead of doing a full `window.location.reload()`. The `<select>` in `AppShell` is replaced by an explicit-click dropdown menu; if a mutation is in flight when you try to switch, `useConfirm` asks first.
- Every TanStack query key is prefixed with `["ws", workspaceId, …]` via the new `wsKey()` helper in `web/src/lib/queryKeys.ts`. Every bare `qc.invalidateQueries()` (no key) is replaced with a workspace-scoped prefix — those used to nuke the entire cache.
- DataTable hardening: selection set prunes when rows change and resets on `tableId` change; CSV export gets a UTF-8 BOM, CRLF terminators, formula-injection neutralisation (`=`, `+`, `-`, `@`, TAB, CR), and `URL.revokeObjectURL()` after the click. New `web/src/components/DataTable.test.tsx` pins the helpers (14 tests).
- `<QueryStateBoundary>` renders a "couldn't load — retry" banner instead of a silent empty list on Parts / Orders / Builds / Projects / Lots.
- PartsList bulk-delete now goes through `useConfirm` (no more hand-rolled overlay).
- Scanner panels swap stale `bg-bg-soft` / `text-text-muted` classes for the existing `bg-panel2` / `text-muted` tokens (Scanner.tsx + ZxingScanner.tsx).
- `lib/api.ts` now accepts an optional `AbortSignal`; `AuthProvider` aborts the bootstrap `/auth/me` on unmount. Stale-closure bug in the bootstrap effect fixed via a ref.

## What this PR closes

| ID | What |
|---|---|
| FE2-001 | global 401 handler |
| FE2-002 | workspace switcher needs explicit click + confirm |
| FE2-003 | workspace switch flushes cache + drops `window.location.reload` |
| FE2-004 | every query key includes workspaceId |
| FE2-005 | PartsList bulk-delete via `useConfirm` |
| FE2-007 | DataTable selection prune + reset on tableId change |
| FE2-008 | DataTable CSV: BOM, CRLF, formula-injection mitigation, revokeObjectURL |
| FE2-009 | Scanner Tailwind tokens (`bg-bg-soft` → `bg-panel2`, `text-text-muted` → `text-muted`) |
| FE2-010 | login deep-link preservation |
| v1 FE HIGH-1 | auth bootstrap stale closure → ref |
| v1 FE HIGH-4 | api.ts AbortSignal + unmount abort |
| v1 FE HIGH-7 | bare `qc.invalidateQueries()` replaced with key-prefixed calls |
| v1 FE HIGH-8 | query keys workspace-scoped |

## Test plan

- [x] `cd web && npm run build` — `tsc -b` + `vite build` both green
- [x] `cd web && npm test` — 33/33 pass (14 new DataTable tests)
- [ ] Manual: `/api/me` returns 401 → user redirected to `/login` with deep-link preserved → after sign-in lands back on the original URL
- [ ] Manual: open scanner → switch workspace → scanner re-fetches license key (cache flushed)
- [ ] Manual: PartsList bulk-select 5 rows, click Delete → ConfirmDialog appears with the part names → confirm archives
- [ ] Manual: export a list to CSV → opens in Excel without mojibake; a row whose first char is `=` is rendered as text, not evaluated

### DIFF
--- a/web/src/App.tsx
+++ b/web/src/App.tsx
@@ -1,5 +1,5 @@
 import { lazy, Suspense } from "react";
-import { Navigate, Outlet, Route, Routes } from "react-router-dom";
+import { Navigate, Outlet, Route, Routes, useLocation } from "react-router-dom";
 import { AuthProvider, useAuth } from "@/lib/auth";
 import AppShell from "@/components/layout/AppShell";
 import { ConfirmDialogProvider } from "@/components/ConfirmDialog";
@@ -113,8 +113,12 @@ const WorkspaceSettings = lazy(() => import("@/routes/settings/Workspace"));
  */
 function Gate() {
   const { me, loading } = useAuth();
+  const location = useLocation();
   if (loading) return <div className="p-6 text-muted">Loading…</div>;
-  if (!me) return <Navigate to="/login" replace />;
+  // Preserve the deep-link target across the login round-trip so a user
+  // who hits /parts/abc123 with no session lands back there after
+  // signing in (FE2-010).
+  if (!me) return <Navigate to="/login" replace state={{ from: location }} />;
   return (
     <AppShell>
       <Outlet />
--- a/web/src/components/ActivityTimeline.tsx
+++ b/web/src/components/ActivityTimeline.tsx
@@ -14,7 +14,7 @@ import {
 } from "lucide-react";
 import type { ReactNode } from "react";
 import { api } from "@/lib/api";
-import { wsKey } from "@/lib/queryKeys";
+import { useWsKey } from "@/lib/queryKeys";
 
 type ActivityKind =
   | "stock"
@@ -137,7 +137,7 @@ function summary(e: ActivityEntry): string {
 
 export default function ActivityTimeline({ endpoint }: Props) {
   const { data, isLoading } = useQuery({
-    queryKey: wsKey("activity", endpoint),
+    queryKey: useWsKey("activity", endpoint),
     queryFn: () => api.get<ActivityEntry[]>(endpoint),
   });
 
--- a/web/src/components/ActivityTimeline.tsx
+++ b/web/src/components/ActivityTimeline.tsx
@@ -14,6 +14,7 @@ import {
 } from "lucide-react";
 import type { ReactNode } from "react";
 import { api } from "@/lib/api";
+import { wsKey } from "@/lib/queryKeys";
 
 type ActivityKind =
   | "stock"
@@ -136,7 +137,7 @@ function summary(e: ActivityEntry): string {
 
 export default function ActivityTimeline({ endpoint }: Props) {
   const { data, isLoading } = useQuery({
-    queryKey: ["activity", endpoint],
+    queryKey: wsKey("activity", endpoint),
     queryFn: () => api.get<ActivityEntry[]>(endpoint),
   });
 
--- a/web/src/components/AttachmentsPanel.tsx
+++ b/web/src/components/AttachmentsPanel.tsx
@@ -3,7 +3,7 @@ import { useQuery, useQueryClient } from "@tanstack/react-query";
 import { toast } from "sonner";
 import { Trash2, Download, UploadCloud, Paperclip } from "lucide-react";
 import { api, ApiError } from "@/lib/api";
-import { wsKey } from "@/lib/queryKeys";
+import { useWsKey } from "@/lib/queryKeys";
 import { useConfirm } from "@/components/ConfirmDialog";
 
 type Attachment = {
@@ -33,7 +33,7 @@ function humanSize(bytes: number): string {
 export default function AttachmentsPanel({ objectType, objectId, canWrite }: Props) {
   const confirm = useConfirm();
   const qc = useQueryClient();
-  const queryKey = wsKey("attachments", objectType, objectId);
+  const queryKey = useWsKey("attachments", objectType, objectId);
   const { data, isLoading } = useQuery({
     queryKey,
     queryFn: () =>
--- a/web/src/components/AttachmentsPanel.tsx
+++ b/web/src/components/AttachmentsPanel.tsx
@@ -3,6 +3,7 @@ import { useQuery, useQueryClient } from "@tanstack/react-query";
 import { toast } from "sonner";
 import { Trash2, Download, UploadCloud, Paperclip } from "lucide-react";
 import { api, ApiError } from "@/lib/api";
+import { wsKey } from "@/lib/queryKeys";
 import { useConfirm } from "@/components/ConfirmDialog";
 
 type Attachment = {
@@ -32,7 +33,7 @@ function humanSize(bytes: number): string {
 export default function AttachmentsPanel({ objectType, objectId, canWrite }: Props) {
   const confirm = useConfirm();
   const qc = useQueryClient();
-  const queryKey = ["attachments", objectType, objectId];
+  const queryKey = wsKey("attachments", objectType, objectId);
   const { data, isLoading } = useQuery({
     queryKey,
     queryFn: () =>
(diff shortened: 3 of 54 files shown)


matescb (commented):
## Request changes

**Claims to close:** `FE2-001`, `FE2-002`, `FE2-003`, `FE2-004`, `FE2-005`, `FE2-007`, `FE2-008`, `FE2-009`, `FE2-010`, plus v1 `FE HIGH-1/4/7/8` — *but a Rules-of-Hooks regression makes the cache layer ship-broken (see Blockers).*

### What the patch does
- New `web/src/lib/queryKeys.ts` introduces `wsKey()` (hook variant), `wsKeyOf(workspaceId, …)` (vanilla), `wsScope()`, and a tiny `authBus` pub/sub for cross-cutting auth events.
- `main.tsx` wires a `QueryCache`/`MutationCache` `onError` that emits `authBus.emit("unauthorized")` on any `ApiError(401)`. `auth.tsx` translates the event into `nav("/login", { state: { from: location } })` and `qc.clear()`. `Login.tsx`/`Signup.tsx` replay `state.from` after success.
- `AuthProvider.refresh()` now reads `workspaceId` via a ref to dodge the v1 FE HIGH-1 stale closure; bootstrap `/auth/me` is cancellable via `AbortController`. `lib/api.ts` threads an optional `signal` through every wrapper.
- `AppShell.tsx` replaces the `<select>` workspace switcher with a click-only menu and `useConfirm`s if `useIsMutating() > 0`. `switchWorkspace()` drops the full-page reload in favour of `qc.clear()` + `nav("/parts")`.
- `DataTable.tsx` adds selection prune on `rows`/`tableId` change, extracts `escapeCsvCell`/`buildCsv`/`pruneSelection` for unit tests, and produces RFC-4180 + UTF-8 BOM + CRLF + formula-injection-neutralised CSV with `URL.revokeObjectURL`.
- `QueryStateBoundary.tsx` shows a "couldn't load — retry" banner; wired into Parts/Orders/Builds/Projects/Lots lists.
- `PartsList.tsx` bulk-delete migrates to `useConfirm`. Scanner panels swap stale tokens for `bg-panel2`/`text-muted`. ~50 query callsites prefixed with `wsKey(...)`.

### Strengths
- ✅ Clean separation of `wsKey` (hook) vs `wsKeyOf` (vanilla) vs `wsScope` (workspace prefix). The `authBus` pattern is the right way to bridge from `QueryCache.onError` (no React context) into the router.
- ✅ `DataTable.test.tsx` is a tight, deterministic spec for the CSV helpers and `pruneSelection` — formula-injection cases and the "string `-5` vs numeric `-5`" distinction are exactly the right pins.
- ✅ 401 handler correctly avoids redirect loops on `/login`/`/signup` (`auth.tsx:139`) and 401 is filtered out in `QueryStateBoundary` (`QueryStateBoundary.tsx:34`) so the user doesn't see a flash of "couldn't load" mid-bounce.
- ✅ `Login.tsx:21`/`Signup.tsx` deep-link replay works whether the user landed there from `<Gate>` or from a 401 bounce, because both paths set `state.from`.
- ✅ Mutation-aware workspace switcher (`AppShell.tsx:140-149`) is a nice UX touch — confirms before nuking unsaved work.

### Blockers

🐛 **Rules of Hooks violation in `wsKey()` — every mutation handler in the PR will throw "Invalid hook call" at runtime.**

`web/src/lib/queryKeys.ts:39` defines `wsKey()` as a function that calls `useAuth()` (i.e. `useContext`). That's fine when called from a render body (the `useQuery` callsites are correct), but the PR also calls `wsKey(...)` *inside* event handlers passed to `qc.invalidateQueries({ queryKey: wsKey(...) })`. Those handlers run after render, outside React's dispatcher window — `useContext` returns `null` and the `.useContext` access throws. Reproduced in a real React 18 + jsdom harness; clicking the button reliably fires `Invalid hook call. Hooks can only be called inside of the body of a function component.` and trips `Sentry.ErrorBoundary` in `main.tsx`.

Affected callsites (~33 lines, 17 files): `BuildDetail.tsx:91-92,105-106`; `OrderDetail.tsx:64,74,103-104,118-119`; `PartsList.tsx:54`; `ScanImport.tsx:268-269`; `PartAddStock.tsx:46-47`; `PartInfo.tsx:84-85`; `PartMembers.tsx:29,36`; `PartMoveStock.tsx:98-99`; `PartRemoveStock.tsx:95-96`; `PartSettings.tsx:36`; `PartSourcing.tsx:48-49`; `PartSubstitutes.tsx:22,26`; `ProjectBOM.tsx:25,31`; `ProjectData.tsx:16`; `ProjectImport.tsx:186`; `Workspace.tsx:85`; `StorageDetail.tsx:122`. Net result: bulk-delete, archive, restore, add-stock, move-stock, receive-order, build-consume, custom-field edits, project import, BOM edits, member add, substitute add, storage settings toggle — all 500-error the UI on click.

The fix is the pattern you already use in `PartOther.tsx:13,22,27` and `Reports.tsx:77`: capture `const { workspaceId } = useAuth()` at the top of the component and call `wsKeyOf(workspaceId, ...)` (or `wsScope(workspaceId)`) in handlers. Worth either renaming `wsKey` → `useWsKey` to make the hook nature loud, or just deleting the hook variant and standardising on `wsKeyOf` everywhere.

🐛 **CI didn't catch this because the new tests are pure-function only.** `DataTable.test.tsx` doesn't render the component, and there's no other rendering test in the suite. Before re-merging, please add at least one RTL test that mounts `OrderDetail` (or any of the affected files) inside a `QueryClientProvider` + `MemoryRouter` + `AuthCtx.Provider` and clicks an action button — that would have caught this. Manual smoke of "click Archive on a part" is the minimum bar.

### Notes (non-blocking)
- ⚠️ `wsKey()`'s `?? "none"` fallback (`queryKeys.ts:40`) silently scopes pre-bootstrap renders to a shared "none" bucket. Low risk since `<Gate>` blocks render until `me` resolves, but if any non-Gated page ever fires a query (e.g. a future public page) it'd cross-pollinate caches. Worth either enforcing `enabled: !!workspaceId` at the call site or asserting in dev mode.
- ⚠️ `authBus` listeners use `try/catch` around each callback (`queryKeys.ts:65-69`) but swallow errors silently. Consider `console.error`-ing in dev so a buggy listener isn't invisible.
- ⚠️ `escapeCsvCell` (`DataTable.tsx:32`) only neutralises *string* cells — but a `Number` whose source was a user-supplied free-text field that happened to parse cleanly will round-trip unquoted. Probably fine in practice; pin in a comment if so.
- ⚠️ The 401 handler doesn't cancel in-flight mutations before redirecting (`auth.tsx:147-150`). A long-running PATCH that 401s then 201s after the cookie was cleared would land a write under the next user's session if they share the browser. Edge-case but worth a `qc.cancelQueries()` + `qc.getMutationCache().clear()` before the nav.
- ⚠️ `AppShell.tsx:140` uses `useIsMutating()` which returns the count of *all* in-flight mutations, including background ones the user didn't initiate (e.g. an autosave from another tab via shared cache). Confirm prompt may fire spuriously. Probably acceptable.
- ⚠️ `PartsList.tsx:36-44` — the confirm message embeds part names verbatim. Names are user-controlled but the dialog is plain text so no injection risk; just noting.

### Ops notes

None — pure FE. But because this is the cache + auth layer and there's no staging, **do not merge** until the hook bug is fixed and a smoke test is added; a destructive merge here means the prod app crashes on every "Archive part", "Move stock", "Receive order", etc., and the user has to hard-refresh.

PR #34, opened 2026-05-02
fix(backend): correctness + tx boundary + isolation/RBAC tests (BE2-005/007/009/010/016 + TEST-001/002)

matescb:
## Summary

Closes the v2 teardown's remaining backend-correctness items + extends the test matrix to the routers and roles that were uncovered.

### IDs closed

| ID | What |
|---|---|
| `BE2-005` | 5 ad-hoc per-part / per-lot `SUM(quantity_delta) GROUP BY` aggregations now route through new `bulk_current_quantities` / `bulk_current_quantities_by_lot` helpers in `domain/stock/service.py`. The "current = SUM(delta)" invariant has exactly one expression in code. |
| `BE2-007` | `move_stock` pre-assigns the OUT/IN UUIDs so both `StockEntry` rows can carry their cross-reference at insert time. One `add_all + flush` instead of three flushes-and-mutate. The lot-split path is wrapped in `db.begin_nested()` so a downstream raise cleans up the orphan lot. |
| `BE2-009` | New `require_resource_access` helper in `api/_helpers.py` does existence -> workspace-match -> role-check in that order. Archive/restore on parts / projects / orders / builds / storage migrated off `dependencies=[Depends(require_role("admin"))]`. A non-admin probing a foreign workspace's id now gets 404, not 403 -- the membership oracle is closed. |
| `BE2-010` | `get_db` now commits on clean route exit and rolls back on any raised exception (see "Behaviour change" below). |
| `BE2-016` | `_get_part(include_archived=False)` is the default; archived parts refuse PATCH / substitute add / BOM add/patch/match. Read paths pass `include_archived=True` so the detail page + activity timeline still load. |
| `TEST-001` | `test_workspace_isolation.py` extended with lots / orders / invitations / bom_presets / reports / search / parts_provider / catalog / part_activity. |
| `TEST-002` | New `test_rbac_member.py` mirrors the viewer matrix; covers member-can-write + member-cannot-archive/restore/bulk-delete + member-cannot-PATCH-workspace-secrets. Adds attachments + custom_fields coverage missing from the viewer matrix. |

### Behaviour change reviewers must inspect: `db.commit()` sweep

`get_db()` now commits on yield-success and rolls back on raise. Every route handler that previously called `db.commit()` had it removed -- see the diff for the full list. Routes touched:

- `parts.py` (create_part, patch_part, archive_part, restore_part, bulk_delete_parts, add_substitute, del_substitute, add_member, del_member, refresh_from_provider, quick_remove_bag)
- `projects.py` (create_project, patch_project, archive_project, restore_project, add_entry, patch_entry, del_entry, commit_bom, match_entry)
- `orders.py` (create_order, patch_order, archive_order, restore_order, add_entry, patch_entry, del_entry, receive_order)
- `builds.py` (create_build, patch_build, archive_build, restore_build, consume_build)
- `storage.py` (create_storage, patch_storage, archive_storage, restore_storage)
- `stock.py` (add, remove, move, adjust)
- `lots.py` (patch_lot, move_lot, adjust_lot)
- `bom_presets.py` (create_preset, patch_preset, delete_preset)
- `tags.py` (create, link, unlink)
- `attachments.py` (upload, delete)
- `custom_fields.py` (create_or_update, delete, restore_override)

**Single load-bearing exception**: `bulk_import_from_scan` keeps its terminal `db.commit()`. Per-row savepoints (`db.begin_nested()`) accumulate into the outer transaction; the explicit commit ensures every accumulated savepoint is durable BEFORE we serialise the response. The dep's clean-exit commit then becomes a no-op. The decision and the comment are in `parts.py:1013-1021`.

**Files explicitly NOT touched** (other batches own them): `auth.py`, `workspaces.py`, `invitations.py`. Their `db.commit()` calls remain in place.

### Things to look at carefully

- `receive_order` / `consume_build` previously had `try / except / db.rollback() / raise` shape. Removed because `get_db` rolls back on raise -- but reviewers should confirm the "raise inside the try" path still surfaces a 4xx (it does: `raise HTTPException(...)` propagates and the dep finalizer rolls back).
- `move_stock` lot-split now uses `db.begin_nested()`. The savepoint context manager is inside an outer transaction the dep manages -- nesting works as expected with SQLAlchemy's "future" mode, but worth a second pair of eyes.
- The new `require_resource_access` helper is intentionally NOT a FastAPI dependency: it's a regular function the route calls inline. This keeps it generic over Model + path-param-name without dynamic-signature gymnastics.

## Test plan

- [ ] CI green (backend pytest + tsc -b + vite build)
- [ ] `test_db_tx.py`: rollback-on-raise actually rolls back through the dep
- [ ] `test_archive_oracle.py`: foreign workspace id -> 404, same-ws non-admin -> 403 on parts / projects / orders / builds / storage
- [ ] `test_archived_part_bindings.py`: writes refuse archived parts; reads still work
- [ ] `test_move_stock.py`: both entries land with cross-references; lot-split creates child lot
- [ ] `test_rbac_member.py`: member can write non-admin endpoints; admin-only endpoints reject with 403
- [ ] `test_workspace_isolation.py` extended assertions: 404 across workspaces for lots / orders / bom_presets / reports / search results / parts_provider / catalog
- [ ] Smoke: a normal create-part / add-stock / move-stock / patch-part round-trip in dev still 200s with no explicit `db.commit()` left in the routes

### DIFF
--- a/backend/app/api/_helpers.py
+++ b/backend/app/api/_helpers.py
@@ -7,6 +7,8 @@ from uuid import UUID
 from fastapi import HTTPException, status
 from sqlalchemy import select
 
+from app.core.deps import _ROLE_RANK, _membership_role
+
 
 def assert_in_workspace(
     db,
@@ -76,3 +78,71 @@ def assert_polymorphic_in_workspace(
             detail=f"unknown object_type: {object_type}",
         )
     return assert_in_workspace(db, Model, object_id, workspace_id, label=object_type)
+
+
+def require_resource_access(
+    db: Any,
+    Model: Any,
+    id_: UUID,
+    *,
+    user: Any,
+    ws: Any,
+    role: str = "member",
+    label: str | None = None,
+) -> Any:
+    """Resolve a workspace-owned resource by id and gate on the caller's
+    role — in the right order so the response status leaks no info.
+
+    The order is the whole point (BE2-009):
+
+      1. existence in the DB
+      2. membership in the caller's current workspace
+      3. role check for that workspace
+
+    Failing (1) or (2) → 404. The caller is told nothing about whether
+    the row exists; the response is identical for "no such id" and "id
+    in another workspace". Failing (3) → 403: the row exists *in this
+    workspace*, the caller just lacks the named role for it.
+
+    The previous shape used `dependencies=[Depends(require_role("admin"))]`,
+    which runs the role check BEFORE the per-resource lookup. A member
+    probing an archive endpoint with a foreign workspace's UUID got 403
+    — an oracle telling the prober "this UUID exists somewhere; you
+    just lack the role". Resource-first + role-second closes that.
+
+    Usage::
+
+        @router.post("/{part_id}/archive")
+        def archive(part_id: UUID, db: DbSession, ws: CurrentWorkspace, user: CurrentUser):
+            p = require_resource_access(db, Part, part_id, ws=ws, user=user, role="admin")
+            ...
+
+    Generic over `Model` — works for Part / Project / Order / Build /
+    StorageLocation. The 404 message uses `label` (or the model's
+    tablename) so the client gets a recognisable string.
+    """
+    floor = _ROLE_RANK[role]
+    name = label or getattr(Model, "__tablename__", "row")
+
+    # 1) existence + 2) workspace match — both fold into a single 404.
+    # We call `db.get` (unscoped) and follow with the workspace equality
+    # check; the 404 branch covers both "no such row" and "row in
+    # another workspace" with the same response so a foreign-id probe
+    # can't distinguish them.
+    row = db.get(Model, id_)
+    if row is None or getattr(row, "workspace_id", None) != ws.id:
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND,
+            detail=f"{name} not found",
+        )
+    # 3) role check on the (now-known-to-be-in-workspace) resource. 403
+    # here is correct — it tells the caller "you exist in this
+    # workspace, the row exists, but you lack `role`" without leaking
+    # anything about other workspaces.
+    rank = _ROLE_RANK.get(_membership_role(db, user, ws), 0)
+    if rank < floor:
+        raise HTTPException(
+            status_code=status.HTTP_403_FORBIDDEN,
+            detail=f"requires role {role}+",
+        )
+    return row
--- a/backend/app/api/routes/attachments.py
+++ b/backend/app/api/routes/attachments.py
@@ -170,7 +170,7 @@ async def upload(
         updated_by=user.id,
     )
     db.add(a)
-    db.commit()
+    db.flush()
     return ok(_serialize(a))
 
 
@@ -223,5 +223,4 @@ def delete(attachment_id: UUID, db: DbSession, ws: CurrentWorkspace):
     except FileNotFoundError:
         pass
     db.delete(a)
-    db.commit()
     return ok(None, "deleted")
--- a/backend/app/api/routes/auth.py
+++ b/backend/app/api/routes/auth.py
@@ -82,7 +82,6 @@ def signup(request: Request, payload: SignupIn, response: Response, db: DbSessio
 
     sess = create_session_row(db, user.id)
     user.last_login_at = datetime.now(timezone.utc)
-    db.commit()
 
     _set_session_cookie(response, sess.token)
     log.info("signup", extra={"user_id": str(user.id), "workspace_id": str(ws.id)})
@@ -102,7 +101,6 @@ def login(request: Request, payload: LoginIn, response: Response, db: DbSession)
         raise HTTPException(status_code=status.HTTP_401_UNAUTHORIZED, detail="invalid credentials")
     sess = create_session_row(db, user.id)
     user.last_login_at = datetime.now(timezone.utc)
-    db.commit()
     _set_session_cookie(response, sess.token)
     log.info("login", extra={"user_id": str(user.id)})
     return ok({"user": {"id": str(user.id), "email": user.email, "name": user.name}})
@@ -114,7 +112,6 @@ def logout(request: Request, response: Response, db: DbSession):
     token = request.cookies.get(cookie_name)
     if token:
         revoke_session(db, token)
-        db.commit()
     response.delete_cookie(cookie_name, path="/")
     log.info("logout")
     return ok(None, "logged out")
--- a/backend/app/api/routes/bom_presets.py
+++ b/backend/app/api/routes/bom_presets.py
@@ -66,7 +66,7 @@ def create_preset(payload: PresetIn, db: DbSession, ws: CurrentWorkspace, user: 
         updated_by=user.id,
     )
     db.add(p)
-    db.commit()
+    db.flush()
     return ok(_serialize(p))
 
 
@@ -90,7 +90,6 @@ def patch_preset(preset_id: UUID, payload: PresetPatch, db: DbSession, ws: Curre
     if payload.config is not None:
         p.config_json = json.dumps(payload.config)
     p.updated_by = user.id
-    db.commit()
     return ok(_serialize(p))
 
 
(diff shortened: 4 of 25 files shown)


matescb (commented):
## Approve with notes

**Closes:** `BE2-005`, `BE2-007`, `BE2-009`, `BE2-010` (partial), `BE2-016`, `TEST-001`, `TEST-002`.

### What the patch does

- **BE2-005** — Adds `bulk_current_quantities()` and `bulk_current_quantities_by_lot()` in `backend/app/domain/stock/service.py:146-211`. Both `reports.py::low_stock`, `stock_value`, and `expiring_lots` are rewritten to funnel through them, removing four ad-hoc `SELECT … GROUP BY part_id/lot_id` blocks. The "current = SUM(delta)" invariant now has exactly one expression per axis (part / lot) — the CLAUDE.md "no qty column" invariant is now actually enforceable.
- **BE2-007** — `move_stock` (`backend/app/domain/stock/service.py:489-614`) pre-assigns both `out_id`/`in_id` UUIDs and wraps the writes in `db.begin_nested()`. The IN row carries the back-pointer at insert time; the OUT row's back-pointer is patched in a third flush inside the same savepoint. Split-lot creation is wrapped in the same savepoint so a downstream raise can't orphan the new `Lot`.
- **BE2-009** — New `require_resource_access()` helper in `backend/app/api/_helpers.py:82-148`. Existence + workspace match → 404 first; role check → 403 second. Replaces the `dependencies=[Depends(require_role("admin"))]` shape on every archive/restore endpoint (parts, orders, projects, builds, storage). Foreign-id probes now collapse to the same 404 as "no such row."
- **BE2-010 (the transaction sweep)** — `get_db()` (`backend/app/infra/db.py:19-42`) now `commit`s on clean exit and `rollback`s on raise. ~30 explicit `db.commit()` calls were dropped from routes; many `try/except: db.rollback(); raise` shapes collapsed to plain raises.
- **BE2-016** — `_get_part()` gains `include_archived: bool = False` (default refuses archived). All write paths (`PATCH`, substitute add, BOM `add_entry`/`patch_entry`/`match_entry`) get the strict default; reads (`GET`, `/stock`, `/lots`, `/activity`) and removal paths pass `include_archived=True`. New `_assert_part_live()` helper in `projects.py:145-152` mirrors the same 404-on-archived semantics for BOM binds.
- **TEST-001** — `test_workspace_isolation.py` grew from one test to 25, covering attachments, custom_fields, tags, stock add/remove/move/adjust, parts default-storage / bulk-import, builds, lots, orders, invitations, bom_presets, reports, search, parts_provider, catalog, and activity. Matches the routers the teardown called out.
- **TEST-002** — `test_rbac_member.py` (215 lines, 15 tests) inverts the viewer matrix: member can do all writes, cannot archive/restore/bulk-delete on parts/orders/projects/storage/builds, cannot patch workspace secrets.
- New `test_db_tx.py` (101 lines) mounts a synthetic test-only router on `app` and pins both halves of the dep contract — clean exit commits, raise rolls back.

### Strengths

- ✅ Test deltas alone are 938 LOC for ~620 LOC of production change. Coverage matches the bug surface.
- ✅ The `require_resource_access` docstring (`_helpers.py:88-122`) explains *why* the existence-then-role ordering matters with a worked oracle example. Future contributors won't undo it.
- ✅ `bulk_current_quantities*` returns `dict[UUID, int]` with "missing → 0" convention documented in the docstring; eliminates N+1 from PR #28's hot path.
- ✅ The dep-rollback rewrite is genuinely uniform — the `try/except: db.rollback(); raise` pattern is now expressed as a single bare raise everywhere except `bulk_import_from_scan`, where the per-row savepoints justify the explicit commit.
- ✅ The `get_db` docstring (`infra/db.py:20-32`) explicitly calls out `bulk_import_from_scan` as the one allowed exception. Readers won't flail wondering whether they should add commits.

### Notes (non-blocking)

- ⚠️ **`accept_invitation` was named in BE2-010 and was not refactored.** `backend/app/api/routes/invitations.py:118,147,210`, `auth.py:85,105,117`, and `workspaces.py:49,145,213,227` still hold explicit `db.commit()` calls. With the new `get_db` the duplicate commit is harmless (the second is a no-op), but the PR description's claim that BE2-010 fixes `accept_invitation` is technically only true because the route's commit happens to be the last write before the response — the route shape that the ticket flagged is unchanged. Either rip the explicit commits out of those three files for consistency, or amend the PR description.
- ⚠️ **Stale comment in `move_stock`.** `service.py:497-501` says "One `db.add_all([...]); db.flush()` after this assignment lands both rows atomically with consistent back-pointers." The follow-up commit `741455b` reverted to a three-step write under savepoint (because `related_entry_id` FK is non-deferrable), which the rest of the function correctly reflects. The intro comment now lies about what the code does. Trim those 5 lines and let the per-block comments at `:548` and `:583` carry the explanation.
- ⚠️ **`bulk_import_from_scan` justification is shaky.** `parts.py:1015-1022` says the explicit commit is needed so accumulated savepoints are "durable BEFORE we serialise out_rows." Savepoint releases aren't independently durable — they only become durable when the outer transaction commits. The real reason is "we want to commit the batch even if something later in the response build raised, so a partial batch isn't lost." The current comment will mislead the next reader; either fix it or drop the explicit `db.commit()` and trust the dep.
- ⚠️ **Cross-package `_`-prefixed import.** `_helpers.py:10` imports `_ROLE_RANK` and `_membership_role` from `app.core.deps`. Both are private-by-convention. If you keep this, promote them to public names in `deps.py` (or lift the rank table into `_helpers.py` since that's where it's now consumed).
- ⚠️ **File-system side effects vs the tx boundary.** Routes like `attachments.py::upload` and `parts.py::refresh_from_provider` write to `UPLOAD_DIR` *before* the dep commits. If anything raises between the file write and the response, the asset is orphaned on disk. Pre-existing behaviour, not a regression — but the dep refactor now makes this asymmetry more visible. Worth a follow-up ticket; not a blocker.
- ⚠️ **`bulk_current_quantities*` skips `bucket_match`.** PR #28's reservation locking added a per-bucket lock for `release_reservations`. The new bulk helpers query `StockEntry` directly without that lock — fine for read-only reports, but worth a one-liner in the docstring noting "read-only, not safe for reservation arithmetic."
- ⚠️ `test_move_stock.py:73-91` says the failure-path on lot-split "is harder to provoke from the API alone but is verified by the import + tests passing as a whole." Translation: there's no test that the savepoint actually rolls back a dangling lot. If someone undoes `db.begin_nested()` the test suite stays green. Worth a unit test calling `move_stock()` directly with a forced raise via monkeypatch.

### The db.commit() sweep specifically

I tracked every removed call. Net change: ~30 `db.commit()` and ~10 `db.rollback()` calls deleted across 11 route files; replaced by `db.flush()` at the existing sites where the route still needed the row to be assigned an ID before the response (parts/orders/storage/projects/tags create endpoints). No site lost a flush it actually needed. No `db.commit()` got moved into a code path that previously committed inside a `try` and now commits outside (which would have inverted error semantics). The one `db.commit()` that survives in routed code (`parts.py:1021`) is correctly justified by its position relative to per-row savepoints.

The audit caveat is the one above: `invitations.py`, `auth.py`, `workspaces.py` are entirely untouched. The dep change makes their explicit commits idempotent and the routes happen to commit-then-return, so they're behaviorally fine — but the sweep isn't complete. Suggest a follow-up ticket "remove residual `db.commit()` calls in non-routed write paths" rather than blocking this PR.

### Ops notes

- No migration. No env-var change. No dependency change.
- The dep change means any future route that calls `db.commit()` mid-handler and then raises will lose the side-effect-free part of its work to the dep's rollback (because SA can't rollback a previously-committed segment, but the dep's `except: rollback()` will run on the *current* tx). This is the intended semantics and is documented, but worth flagging in the next dev/architecture refresh that mid-handler commits are now actively wrong, not just unnecessary.
- Per-request rollback on raise means error-response side effects (Sentry tags written to row state, audit-log rows added in middleware) won't survive. Spot-check any middleware that writes to the DB before/after the route.
- Test count delta is significant (~50 new tests). Backend pytest wall time will go up; CI was green so it fits in budget, but worth watching.

LGTM — approve once the `accept_invitation` story and the two stale comments are reconciled (any of: amend PR description, fix comments, or open follow-up tickets and merge).

Issue #92, opened 2026-05-02
[INFRA2-019] nginx `proxy_read_timeout 60s` is too short for BOM imports / scan imports


matescb:
**Source:** [`docs/teardown/infrastructure.md`](https://github.com/matescb/stockManager/blob/main/docs/teardown/infrastructure.md) (v2 teardown) — search for `### INFRA2-019:`

**Area:** infrastructure  •  **Priority:** medium

---

Severity: **Medium**

Evidence:

- `deploy/nginx-web.conf:28` sets `proxy_read_timeout 60s` on the `/api/` location.
- BOM import endpoints (BE) operate on a base64 payload that can encode tens of thousands of rows, with per-row DB writes and provider lookups. `SEC-007` documents the unbounded payload concern; even with a sensible cap, a 5000-row BOM will easily exceed 60s on a small VPS.
- The user's `CHANGELOG.md`-flagged scan-to-import flow is similarly long-tailed.
- Apache's vhost (`deploy/parts.matescb.cz.conf:23`) also has no `ProxyTimeout` directive, defaulting to 60s — same problem one layer up.

Impact:

A medium-sized BOM upload returns 504 Gateway Timeout from nginx (or Apache) while the backend is still happily writing rows. The user doesn't know whether to retry (creating duplicate work) or wait. State is half-applied because there's no idempotency token at the route layer either.

Fix instruction:

Bump `proxy_read_timeout` and Apache's `ProxyTimeout` to 5m for `/api/parts/bom/*` and `/api/scan/*` paths specifically, leave 60s on everything else (use a dedicated `location /api/parts/bom/` block in nginx). Pair with `SEC-007`'s row-cap fix so timeouts can't be used as a "block until OOM" attack.

---

_Auto-pushed from teardown report. Originally identified by Claude Code review pass — see `docs/teardown/SUMMARY.md` for the full priority queue._

matescb:
## Implementation plan

### Context

[...]

PR #137, opened 2026-05-02
fix(auth): index user_sessions.expires_at + periodic purge

### LINKED ISSUE #98: [DB-007] `user_sessions` has no index on `expires_at` and the table grows unbounded
**Source:** [`docs/teardown/database.md`](https://github.com/matescb/stockManager/blob/main/docs/teardown/database.md) (v2 teardown) — search for `### DB-007:`

**Area:** database  •  **Priority:** low

---

Severity: **Medium**

Evidence:

- `backend/alembic/versions/0001_initial.py:33-41` — only indexes `ix_user_sessions_user_id`.
- `backend/app/domain/users/models.py:30-35` — `expires_at` column, no index, no TTL.

Impact:

Any "delete expired sessions" sweep is a full table scan and grows
linearly with logins. There is also no documented cleanup path —
sessions only get deleted on explicit logout. A long-running prod
instance will accumulate every expired session forever.

Fix instruction:

Add a migration that creates `ix_user_sessions_expires_at`. Add a
periodic cleanup task (cron / startup hook) that does `DELETE FROM
user_sessions WHERE expires_at < now()`. Cross-ref Sec issue SEC-006:
sessions are stored as plaintext tokens, which is independently a
problem. If you fix that and re-issue, the cleanup window for the old
plaintext token rows wants this index.

---

matescb:
Closes #98.

Per the auto-generated plan (DB-007 / 2026-05-02). Closes the
unbounded-growth gap on `user_sessions`: prior to this PR the only
index was `ix_user_sessions_user_id`, so `DELETE WHERE expires_at <
now()` was a full table scan and there was no scheduled cleanup.

- Alembic 0019 creates `ix_user_sessions_expires_at`.
- `app/core/auth.py::purge_expired_sessions(db, *, now=None) -> int`
  is the new helper. Idempotent.
- `app/main.py` gains an async `lifespan` context manager that spawns
  a background task running the purge every
  `SESSION_PURGE_INTERVAL_SECONDS` (default 3600). The task uses a
  fresh `SessionLocal()` each tick — connection isn't held across the
  sleep. Cancelled cleanly on shutdown.
- New tests in `backend/tests/test_session_cleanup.py`.

## Test plan
- [ ] CI green (pytest covers the helper + the index existence)
- [ ] Manual: `EXPLAIN DELETE FROM user_sessions WHERE expires_at < now()`
      shows an index scan after this lands
- [ ] Manual: backend startup logs include
      `purge_expired_sessions: scheduled every 3600s`

## Ops notes
- The lifespan hook is the first long-lived background task in the
  app. Verify `--workers 1` (CLAUDE.md invariant) before promoting
  enforcement; multi-worker would run the DELETE from every worker
  (still correct but wasteful).
- `SESSION_PURGE_INTERVAL_SECONDS=0` disables the scheduled task; the
  index still helps a manual / future-cron sweep.

## Coordination
Migration is numbered 0019 because main is at 0018 at the time of
authoring. **Several adjacent open PRs may also want 0019** (per the
locked-constraints document). If any of them lands before this one,
rebase onto the new head and bump both the filename and
`revision = "..."` to match. Touches `app/main.py` like #92 (proxy
read timeout) but in a different region — no conflict expected.

### DIFF
--- a/backend/alembic/versions/0019_user_sessions_expires_at_idx.py
+++ b/backend/alembic/versions/0019_user_sessions_expires_at_idx.py
@@ -1,0 +1,49 @@
+"""Index user_sessions.expires_at + give the periodic cleanup task a
+seekable column.
+
+Revision ID: 0019
+Revises: 0018
+Create Date: 2026-05-02
+
+DB-007 / issue #98. Closes the unbounded-growth gap on `user_sessions`:
+prior to this migration the only index was `ix_user_sessions_user_id`,
+so any "delete expired" sweep was a full table scan. The plan was to
+keep these rows around forever — sessions were only deleted on explicit
+logout — so a long-running prod accumulates every expired row.
+
+Pairs with the lifespan-hook purge in `app/main.py` that runs
+`DELETE FROM user_sessions WHERE expires_at < now()` once an hour. The
+DELETE planner picks this index because every row has a non-null
+`expires_at`; no partial predicate is needed.
+
+NOTE on chain: this migration is numbered 0019 because main had 0018
+at the time of authoring. Several open PRs in adjacent batches may
+also want 0019 — if any of them lands first, rebase this onto the new
+head and bump both the filename and `revision = ...` to match. Don't
+edit a migration once it's on `main` (CLAUDE.md invariant).
+"""
+from __future__ import annotations
+
+from alembic import op
+
+
+revision = "0019"
+down_revision = "0018"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.create_index(
+        "ix_user_sessions_expires_at",
+        "user_sessions",
+        ["expires_at"],
+        unique=False,
+    )
+
+
+def downgrade() -> None:
+    op.drop_index(
+        "ix_user_sessions_expires_at",
+        table_name="user_sessions",
+    )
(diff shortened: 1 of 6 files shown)


matescb:

Index-add migration on `user_sessions.expires_at` plus an in-process hourly purge driven by a FastAPI lifespan hook. Verified:

- Migration 0019 chain clean (continues 0018); revision id matches filename; downgrade drops cleanly. `CREATE INDEX` is non-concurrent but `user_sessions` is small (one row per active session) so lock window is negligible — low severity.
- No race with active sessions: purge predicate `expires_at < now()` is exactly the same predicate `core/deps.py::get_current_user` uses to reject — any row the purge would delete is already 401-rejecting on the auth path.
- Background task uses fresh `SessionLocal()` per tick (no held connection across `await`), `synchronize_session=False` (correct for batch DELETE), `CancelledError` propagates for clean shutdown, other exceptions logged without killing the loop.
- Single-worker invariant (CLAUDE.md) explicitly acknowledged in the docstring; `SESSION_PURGE_INTERVAL_SECONDS=0` knob disables the loop while leaving the index in place for any future cron sweep.
- `index=True` deliberately omitted on the model (with comment) so autogenerate doesn't emit a duplicate `CREATE INDEX` later.
- No surface change to envelope, workspace isolation, MPN uniqueness, asset URLs, bag signature, provider catalog, compose command form, session-cookie `secure` gating, or `--workers 1`.

 review complete at 2026-05-02T09:38:00Z

PR #138, opened 2026-05-02
fix(db): partial bag_signature index (DB-008)

### LINKED ISSUE #99: [DB-008] Bag-signature index does not have a partial `WHERE bag_signature IS NOT NULL` predicate
**Source:** [`docs/teardown/database.md`](https://github.com/matescb/stockManager/blob/main/docs/teardown/database.md) (v2 teardown) — search for `### DB-008:`

**Area:** database  •  **Priority:** medium

---

Severity: **Medium**

Evidence:

- `backend/alembic/versions/0012_stock_entries_bag_signature.py:28-32` — `op.create_index("ix_stock_ws_bag_signature", "stock_entries", ["workspace_id", "bag_signature"])` with no predicate.
- The vast majority of `stock_entries` rows have `bag_signature IS NULL` (only scan-import rows set it).

Impact:

The index is bloated with a row per stock entry instead of just the
scan-imported ones (a small fraction). Insert cost on every ledger
write pays for an index that's only read by the scan-recognition
flow. Storage waste at scale, slower writes for the 99% case.

Fix instruction:

Drop and recreate as a partial: `CREATE INDEX
ix_stock_ws_bag_signature ON stock_entries(workspace_id,
bag_signature) WHERE bag_signature IS NOT NULL;`. Update
`stock/models.py:39` to mirror the predicate via
`postgresql_where=text("bag_signature IS NOT NULL")`. Do this
concurrently in prod (`CREATE INDEX CONCURRENTLY` + drop old) — the
ledger is the hottest write table and a non-concurrent reindex blocks
inserts.

---

matescb:
Closes #99.

v2 teardown DB-008. `ix_stock_ws_bag_signature` (added in `0012`) was non-partial — every `stock_entries` row sat in it even though `bag_signature` is only populated on scan-import (a tiny fraction). The ledger is the hottest write table, and a non-partial index pays insert-time cost for every row. New migration `0019` drops + recreates the same index name with `WHERE bag_signature IS NOT NULL`. Both ops use `CREATE INDEX CONCURRENTLY` inside `op.get_context().autocommit_block()` so prod's writes aren't blocked.

Model `__table_args__` Index() now carries `postgresql_where=text("bag_signature IS NOT NULL")` so future autogen sees it as in-sync.

## Test plan
- [ ] CI green (`pytest backend/tests/test_bag_signature_index.py` queries `pg_indexes` for the predicate)
- [ ] Manual on prod after deploy: `SELECT pg_relation_size('ix_stock_ws_bag_signature');` should drop substantially; `EXPLAIN` of a bag-signature lookup should still hit the index.

## Ops notes
- `pg_dump` before deploy (per CLAUDE.md). The migration uses CONCURRENTLY so writes don't block, but a backup is the standard belt-and-braces step for any index ddl on the ledger table.
- If autocommit_block fails on the prod Postgres for any reason, fallback path: apply the SQL manually outside CI window.

## Coordination
None. Out of scope: renaming the index, backfilling historical rows, prod-plan smoke check (ops step).

Generated with Claude Code

### DIFF
--- a/backend/alembic/versions/0019_partial_bag_signature_index.py
+++ b/backend/alembic/versions/0019_partial_bag_signature_index.py
@@ -1,0 +1,58 @@
+"""Partial bag_signature index (DB-008).
+
+Revision ID: 0019
+Revises: 0018
+Create Date: 2026-05-02
+
+The original index from `0012` (`ix_stock_ws_bag_signature` on
+`(workspace_id, bag_signature)`) had no predicate. Almost every
+`stock_entries` row has `bag_signature IS NULL` (only scan-import
+rows ever set it), so the index was bloated and paid an insert-time
+cost on every ledger write for entries that will never be looked up
+by the bag-rescan flow.
+
+This migration drops the existing non-partial index and recreates it
+as a partial index with `WHERE bag_signature IS NOT NULL`. Same name,
+same column ordering — only the predicate changes.
+
+Production note: `stock_entries` is the hottest write table; both
+`DROP INDEX` and `CREATE INDEX` run with `CONCURRENTLY` so writes
+aren't blocked. Concurrent index DDL cannot run inside a transaction,
+so we use alembic's `autocommit_block()` to escape the wrapping
+transaction for these statements only.
+"""
+from alembic import op
+
+
+revision = "0019"
+down_revision = "0018"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # CONCURRENTLY requires running outside any transaction. alembic's
+    # autocommit_block escapes the env.py transaction wrapper for the
+    # statements inside it.
+    with op.get_context().autocommit_block():
+        op.execute(
+            "DROP INDEX CONCURRENTLY IF EXISTS ix_stock_ws_bag_signature"
+        )
+        op.execute(
+            "CREATE INDEX CONCURRENTLY ix_stock_ws_bag_signature "
+            "ON stock_entries (workspace_id, bag_signature) "
+            "WHERE bag_signature IS NOT NULL"
+        )
+
+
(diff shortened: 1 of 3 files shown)


matescb:
Approved by claude-review (no medium+ findings, code-quality CI green).

**Review summary:**
- `alembic-migration-reviewer` checks: `CREATE/DROP INDEX CONCURRENTLY` inside `autocommit_block()` (correct — concurrent DDL can't run in a tx), `IF EXISTS` on the DROP for idempotency, downgrade reverses cleanly to the original non-partial shape, `postgresql_where=text("bag_signature IS NOT NULL")` on the model keeps autogen in sync. No NOT NULL adds, no destructive column ops.
- `/review` invariants: `bag_signature` semantics, column type, and `web/src/lib/bagCode.ts` normalisation untouched. Index still keyed `(workspace_id, bag_signature)` so workspace isolation is preserved at the index level. Scope is strictly DB.
- `/security-review`: no new endpoints, no SQL parameters, no secret surface.
- Minor (non-blocking): there's a brief window between `DROP` and `CREATE CONCURRENTLY` where the `parts.py:506` bag-rescan lookup falls back to a seq scan on `stock_entries`. On the current ledger size this is negligible; if the table grows much larger consider the create-new-then-drop-old pattern in a follow-up.

**Note on the failing `review` check:** the meta `claude-review` GitHub Action failed at startup with `Either ANTHROPIC_API_KEY or CLAUDE_CODE_OAUTH_TOKEN is required` — that's a workflow-secret config issue (the secret isn't wired into the job env), not a defect on this PR. Code-quality checks (`backend-tests`, `web-build`) are green. This review was completed via the local cron-mode equivalent.

 review complete at 2026-05-02T09:39:00Z

PR #161, opened 2026-05-02
fix(types): drop any casts in form-payload assemblers (CQ-004)

matescb:
Closes #120.

The escape hatches `: any` were concentrated where the FE/BE contract lives — request body builders in form pages — so a backend Pydantic change wouldn't surface in `tsc -b`. This PR pins request shapes at the type level without introducing a runtime validator (left for a future zod migration).

## Summary
- `api.post` / `api.patch` (legacy + parsed flavour) now take a `B = unknown` body generic. Untyped callers still compile (TS infers `B = unknown`); typed callers can pass `<TResp, TReqBody>` so a request shape drift is a TS error in the form builder.
- `rawRequest` body parsing typed as `unknown` instead of `any` with a structural narrow on the envelope shape before reaching `.data`.
- `getConflictDetail(err: unknown)` helper — narrows an `ApiError(409)` body to the `{message, existing_id, existing_name}` shape spread by `core/responses.py::http_exception_handler` on MPN collision. `PartCreate` switches to it instead of `e.body as any`.
- Per-form request interfaces mirroring backend Pydantic schemas:
  - `PartCreateRequest` (parts.PartIn)
  - `AddStockRequest` / `RemoveStockRequest` (stock.AddStockIn / RemoveStockIn)
  - `StorageLocationPatchValue` for the storage detail patch field
- Scanner `catch (err: any)` → `catch (err: unknown)` with `instanceof Error` narrowing in `ScanditScanner` + `ZxingScanner`.

No runtime change. Network tab payloads byte-identical before/after.

## Test plan
- [ ] CI green (`npm run build` runs `tsc -b`, `npm test` runs vitest — all 49 currently-existing tests still pass locally)
- [ ] Manual smoke: create a part, add stock, remove stock, edit storage flag — all four flows still POST the same JSON shape (compare network tab before/after)

## Ops notes
None. Pure typing change.

## Coordination
- No file overlap with this batch's other PRs.
- Adjacent to #122 (Pydantic schemas in 4 places) — server-side, no conflict.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #159, opened 2026-05-02
fix(scan): extract ScanImport types module (CQ-003 step 1/4)

matescb:
Closes #119 step 1.

`web/src/routes/parts/ScanImport.tsx` is 731 lines. The plan splits it into co-located sub-components (`ScanImportSession`, `ScanImportRowEditor`, `ScanImportSubmit`) under a folder + `index.tsx` form. This PR is the first step: lift the shared types and tiny utilities into `./ScanImport/types.ts` so the upcoming sub-components can import them without re-declaring.

## Summary
Lifted to `web/src/routes/parts/ScanImport/types.ts`:
- `Row`, `LookupState`, `ImportResultRow`, `ImportResponse` types
- `PROVIDER_LABEL` constant
- `SCAN_IMPORT_SYMBOLOGIES` (the load-bearing module-level symbology constant — inline arrays would tear down the multi-MB scanner SDK on every render; FE CRIT-2)
- `manufacturerMatches` and `newRowId` helpers

`ScanImport.tsx` imports them from `./ScanImport/types` instead of declaring them locally. Zero behaviour change. Same scan-import flow, same dedup keys, same API call shape.

## Remaining steps (deferred to follow-up PRs per the agent-batch's "land FIRST sub-step" rule)
- step 2: extract camera + lookup-with-retry pipeline into `ScanImportSession.tsx`
- step 3: extract `ScanCard` / `BagRescanCard` / `FoundDetails` into `ScanImportRowEditor.tsx`
- step 4: extract storage picker + CTA + summary into `ScanImportSubmit.tsx`, then rename `ScanImport.tsx` → `ScanImport/index.tsx`

## Test plan
- [ ] CI green — `npm run build` (which runs `tsc -b` and Vite bundle)
- [ ] `npm test` — vitest stays green
- [ ] Manual: scan a known bag in dev, observe same UX (camera fires, row pops in, lookup succeeds, "Import (n)" enables, summary banner appears)
- [ ] Manual: re-scan a previously-imported bag, observe `BagRescanCard` still renders

## Ops notes
None. Pure structural refactor, no migration, no API change.

## Coordination
- Coordinates with #118 (parts.py split) — this is the FE side; URL `/api/parts/bulk-import-from-scan` and the `bag_signature: r.bagSig ?? undefined` payload shape stay byte-identical.
- No file overlap with this batch's other PRs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #155, opened 2026-05-02
fix(parts): extract _parts_shared helpers (CQ-002 step 1/4)

matescb:
Closes #118 step 1.

`backend/app/api/routes/parts.py` is 1209 lines (4x the next-largest router) and mixes CRUD, archive, bulk-delete, MPN-uniqueness, asset serving, provider linking, custom-fields, scan-import, bag-signature lookup, and quick-remove-bag. The plan splits it into focused files; this PR is the first step.

## Summary
- New `app/api/routes/_parts_shared.py` lifts the shared helpers (`get_part`, `serialize_part`, `image_urls_for_parts`) and the Pydantic request schemas (`PartIn`, `PartPatch`, `BulkDeleteIn`).
- `parts.py` re-exports them under their existing private names (`_image_urls_for_parts`, `_serialize`, `_get_part`) so call-site references stay byte-identical and the diff is minimal.
- New `tests/test_parts_routing.py` snapshots every `(method, path)` under `/api/parts` so the next split-step PRs can't accidentally drop or rename an endpoint.

No behaviour change. URL surface, response envelope, and dependency chains preserved.

## Remaining steps (deferred to follow-up PRs)
The plan calls for four sub-steps; per the agent-batch's "land FIRST sub-step" rule, only step 1 lands here:
- step 2: extract `/assets/{ws_id}/{filename}` into `parts_assets.py`
- step 3: move `refresh-from-provider` into `parts_provider.py`
- step 4: extract `bulk-delete` + `bulk-import-from-scan` + `by-bag-signature` + `quick-remove-bag` into `parts_bulk.py`

## Test plan
- [ ] CI green (full backend pytest matrix, especially `test_parts_*`, `test_assets`, `test_workspace_isolation`)
- [ ] `tests/test_parts_routing.py` passes — URL surface preserved
- [ ] Manual: `grep -RIn '/api/parts' web/src/` — every URL the FE calls is still mounted at the same path

## Ops notes
None. Pure structural refactor.

## Coordination
- Coordinates with #119 (ScanImport.tsx split) — same scan-import flow but FE side; no file overlap.
- Coordinates with #68 (`bulk_delete_parts` skips other-workspace IDs) — that's a behaviour change; this PR is structure only and #68 should land against the new file once step 4 ships.
- Coordinates with #122 (Pydantic schemas in 4 places) — adjacent but server-side; no conflict.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #152, opened 2026-05-02
fix(invitations): partial composite index + lowercase email

matescb:
Closes #105.

Per the auto-generated plan (DB-014 / 2026-05-02). Adds the partial
composite index that exactly matches the canonical
\"pending invitation for this email in this workspace?\" lookup, and
normalises the application code to RFC 5321 §2.4 case-insensitivity
so the index actually gets used.

- Alembic 0020 creates `ix_invitations_pending_lookup` on
  `(workspace_id, lower(email)) WHERE status = 'pending'`. Backfills
  pre-existing rows to lowercase first (no-op in practice — Pydantic
  EmailStr already normalises on signup).
- `app/api/routes/invitations.py::create_invitation` now lowercases
  payload.email before the membership lookup, the existing-pending
  lookup, and the row insert.
- Two new tests: (a) `Foo@Example.com` then `foo@example.com` hits
  the existing-pending branch on the second call; (b) sanity check
  that `pg_indexes` lists the new index name.
- Two single-column indexes from 0005 stay — they still serve the
  admin-list and accept-invite paths.

## Test plan
- [ ] CI green (new tests cover both paths)
- [ ] `alembic upgrade head` and `alembic downgrade -1` round-trip
      cleanly against a real Postgres
- [ ] Manual: `EXPLAIN ANALYZE SELECT … WHERE workspace_id = 'X' AND
      lower(email) = 'y@z' AND status = 'pending'` shows
      ix_invitations_pending_lookup picked

## Ops notes
- Migration is numbered **0020** because PR #137 (#98 / DB-007)
  reserves 0019 for `ix_user_sessions_expires_at`. **If #137 hasn't
  landed when this PR merges, rebase**: bump filename + revision to
  the next free integer and update `down_revision` accordingly.
  Otherwise alembic will fork on 0018.

## Coordination
Adjacent to #137 (#98) — both add a new migration on top of 0018.
Whoever merges second rebases.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #164, opened 2026-05-02
fix(alembic): rechain duplicate-0019 migrations to 0019 → 0020

matescb:
## Summary

PRs #137 and #138 both landed with `revision = "0019"` / `down_revision = "0018"`. Alembic refuses with "Multiple head revisions" / duplicate revision id. Every PR opened after both merged (~10 in flight) is red on `alembic upgrade head` in CI, and main's own CI is red too (run 25249121114, 25249069040).

Renames `0019_partial_bag_signature_index.py` → `0020_*` and updates its revision/down_revision so the chain is linear:

```
0018 → 0019 (user_sessions.expires_at idx)
     → 0020 (partial bag_signature idx)
```

Both indexes are independent (different tables, different predicates), so the order is arbitrary; user_sessions = 0019 because PR #137 merged first in clock order.

## Why we're editing a merged migration

CLAUDE.md says don't edit migrations on main. The rule exists because doing so breaks the chain on the next deploy. Here, the chain is **already broken** — this PR is the minimal fix that restores it. The actual upgrade SQL on each side is unchanged.

## Test plan

- [ ] CI green on this PR (proves the chain is valid)
- [ ] After merge: PRs in flight (#132, #134, #143, #151, #152, #155, #159, #161, etc.) go green on next CI run
- [ ] Prod deploy: alembic at 0019 (the user_sessions idx, since alembic registered whichever it found first), advances to 0020 = bag_signature partial index. Pre-deploy `pg_dump` runs automatically (PR #29 plumbing).

## Ops notes

The auto-deploy on merge will run `alembic upgrade head` on the running prod container. If alembic's prod state is already at 0019 (likely — both files declared revision="0019" so alembic registered one of them), the upgrade will apply only 0020. If somehow alembic picked the bag-signature 0019 before merge, the user_sessions migration may be missing in prod — out-of-scope to investigate but worth verifying via `\d user_sessions` post-deploy to confirm `ix_user_sessions_expires_at` exists.

### DIFF
--- a/backend/alembic/versions/0020_partial_bag_signature_index.py
+++ b/backend/alembic/versions/0020_partial_bag_signature_index.py
@@ -24,8 +24,8 @@ transaction for these statements only.
 from alembic import op
 
 
-revision = "0019"
-down_revision = "0018"
+revision = "0020"
+down_revision = "0019"
 branch_labels = None
 depends_on = None
 


matescb:
Approved by claude-review (real CI green; unblocks the duplicate-0019 head on main).

PR #177, opened 2026-05-02
fix(infra): container hardening — USER appuser, init service, cap_drop (#85)

### LINKED ISSUE #85: [INFRA2-011] Backend Dockerfile leaves USER unset; runs as root through the chown shim
**Source:** [`docs/teardown/infrastructure.md`](https://github.com/matescb/stockManager/blob/main/docs/teardown/infrastructure.md) (v2 teardown) — search for `### INFRA2-011:`

**Area:** infrastructure  •  **Priority:** medium

---

Severity: **High**

Evidence:

- `backend/Dockerfile:27-35` says "USER is intentionally NOT set here" with the rationale that compose's `command:` runs `chown ... && exec gosu appuser ...` (`docker-compose.prod.yml:87`).
- The chown-then-gosu pattern works only when the compose `command:` is the actual runtime entrypoint. The Dockerfile's default `CMD` (`backend/Dockerfile:35`) does include the chown+gosu shim, so direct `docker run` is also covered — but a future contributor running ad-hoc `docker compose run backend bash` ends up as root inside the container.
- More structurally: this only drops privileges *after* root has already run an arbitrary `chown -R` against a host-bind-mounted path. If `UPLOAD_DIR` is ever pointed at an unintended host path (operator typo, future bind-mount config drift), root in the container will recursively chown a host directory to UID 1000 before dropping. There's no defensive whitelist.
- `cap_drop`, `read_only`, `security_opt: [no-new-privileges]` are not set on any service in `docker-compose.prod.yml`.

Impact:

(1) Container escapes a privilege check away from root on the VPS — defence in depth is one less layer than it should be. (2) Ad-hoc shells end up as root and risk poisoning the bind-mounted `uploads` volume's ownership. (3) No `read_only` root filesystem means a write-anywhere RCE has the run of `/`.

Fix instruction:

Move the chown to a one-time `docker volume`-init job (e.g. a separate compose service that runs once with `restart: no`), then set `USER appuser` in the Dockerfile's runtime stage and drop the gosu trampoline. Add to each service: `security_opt: [no-new-privileges:true]`, `read_only: true` (with explicit `tmpfs:` for `/tmp` and the upload dir), `cap_drop: [ALL]`, `cap_add: [CHOWN, NET_BIND_SERVICE]` only as needed. Verify with `docker inspect <ctr> | jq '.[0].HostConfig.{ReadonlyRootfs,SecurityOpt,CapDrop,CapAdd}'`.

---

matescb:
Closes #85

## Summary
- `backend/Dockerfile`: removes gosu, adds `USER appuser` (UID 1000), simplifies CMD — the runtime process no longer starts as root
- New `backend-init` one-shot service in `docker-compose.prod.yml`: runs `chown -R 1000:1000 /data` as root then exits (`restart: no`); replaces the old gosu trampoline
- `backend` service now `depends_on: backend-init: condition: service_completed_successfully` — clean handoff guaranteed before alembic/uvicorn start
- All three services (`db`, `backend-init`, `backend`, `web`) get `security_opt: ["no-new-privileges:true"]` + `cap_drop: ["ALL"]` with selective `cap_add` per service needs
- `CLAUDE.md` updated with the backend-init pattern in "Things that have bitten us"

Note: `read_only: true` deferred — not applied in this PR to avoid unexpected boot failures from tmpfs requirements in uvicorn/alembic.

## Tests
Backend: 465 passed, 1 skipped, 3 xfailed (no regressions)
Frontend: N/A

## Verify after deploy
- `docker exec backend id` → should show `uid=1000(appuser)`
- `docker compose -f docker-compose.prod.yml ps` → `backend-init` shows `Exited (0)`
- `docker compose -f docker-compose.prod.yml run --user 0 backend bash` → still works for ad-hoc root shell

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -129,6 +129,15 @@ them, that's the bug.
   ran as a separate (failing) shell command. Result: slowapi bucketed
   every client by the docker bridge IP. If you reformat the compose
   file, keep the array form.
+- **`backend-init` one-shot service handles `chown /data` before `backend`
+  starts.** The backend Dockerfile sets `USER appuser` (UID 1000); the
+  runtime container has no root privileges. Ownership of the `uploads`
+  named volume is fixed by `backend-init` (`restart: no`,
+  `command: ["sh","-c","chown -R 1000:1000 /data"]`) which runs as root
+  and exits cleanly. `backend` declares
+  `depends_on: backend-init: condition: service_completed_successfully`
+  so it waits for a clean exit. Don't reintroduce gosu or a root-prefixed
+  `command:` in the backend service — that was the pattern this replaced.
 - **Session cookie `secure` is gated on `APP_ENV == "prod"`** in
   `backend/app/api/routes/auth.py::_set_session_cookie`. Don't make it
   unconditional — local dev runs over HTTP and the cookie wouldn't
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -26,10 +26,11 @@ ENV PYTHONDONTWRITEBYTECODE=1 \
     PIP_NO_CACHE_DIR=1
 
 # libpq5 — psycopg C extension ABI dependency.
-# gosu  — compose command: chowns /data as root then drops to appuser.
 # curl  — healthcheck inside the container hits /api/health.
+# gosu removed — /data ownership is now fixed by the backend-init one-shot
+# service (see docker-compose.prod.yml) before this container starts.
 RUN apt-get update && apt-get install -y --no-install-recommends \
-        libpq5 gosu curl \
+        libpq5 curl \
     && rm -rf /var/lib/apt/lists/* \
     && useradd -r -u 1000 -m -d /home/appuser appuser \
     && mkdir -p /data/uploads \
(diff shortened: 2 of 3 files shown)


matescb:
Rebased onto latest origin/main to resolve merge conflicts.

matescb:

Verified:
- USER appuser (UID 1000) set in Dockerfile; `/data` ownership re-asserted by backend-init service before backend starts (idempotent chown).
- alembic + uvicorn run unprivileged — no disk writes outside `/data/uploads` (which is owned 1000:1000).
- `cap_drop: ALL` on backend with no `cap_add` — appropriate for a Python web app on port 8000. db/web/backend-init cap-add lists are minimal and justified.
- `docker-compose.prod.yml` `command:` stays in single-line JSON-array form (CLAUDE.md invariant preserved).
- `--workers 1`, `--proxy-headers --forwarded-allow-ips=*`, `--timeout-graceful-shutdown 25`, `stop_grace_period: 30s` all preserved.
- No `init: true` introduced (no PID-1/uvicorn shutdown conflict to worry about).

PR #185, opened 2026-05-02
fix(parts): server-side bag_signature recompute + schema tightening (#62)

### LINKED ISSUE #62: [BE2-015] `bag_signature` is recomputed on the client, never re-validated server-side
**Source:** [`docs/teardown/backend.md`](https://github.com/matescb/stockManager/blob/main/docs/teardown/backend.md) (v2 teardown) — search for `### BE2-015:`

**Area:** backend  •  **Priority:** medium

---

Severity: **Medium**

Evidence:
- `backend/app/api/routes/parts.py:443` `find_by_bag_signature` reads `signature: str` and only checks `len == 64 and isalnum`.
- `backend/app/api/routes/parts.py:811` `ScanImportRow.bag_signature: str | None = Field(default=None, max_length=64)`.
- `backend/app/domain/stock/schemas.py:42` `AddStockIn.bag_signature: str | None = Field(default=None, max_length=64)`.
- The canonical normalisation lives only in `web/src/lib/bagCode.ts`. There is no Python-side equivalent.

Impact:

The client computes the signature; the server stores it without verification. A buggy or adversarial client can compute a wrong signature, store it, and the rescan-recognition flow then resolves the wrong bag the next time around — surfacing a stranger's lot to an operator who scanned a different bag. There is also no length=64 hex enforcement on the write side (anywhere up to 64 chars including non-hex passes the schema).

Fix instruction:

Either (a) drop the client-supplied signature and recompute it server-side from the raw bag code (require the raw code in the payload), or (b) add a Python implementation of the same normalisation in `app/domain/parts/services/bag_signature.py` and call it server-side as a verification step. Tighten the schema to `pattern=r'^[a-f0-9]{64}$'`. Add a test that asserts `bag_signature(raw)` (Python) == `bagSignature(raw)` (TypeScript) on a fixture set of real bag codes.

---

matescb:
Closes #62

## Summary
- Add `backend/app/domain/parts/services/bag_signature.py` with `compute_bag_signature` — a Python port of the TS `bagSignature()` normalisation (JS-compatible trim → normalizeControlPictures → trim → SHA-256 hex)
- Tighten `bag_signature` field on `AddStockIn` and `ScanImportRow` from `max_length=64` to `^[a-f0-9]{64}$` regex pattern, blocking adversarial/buggy clients from persisting invalid signatures at schema-validation time
- Add optional `raw_bag_code` to both schemas; when present, server recomputes the digest and rejects mismatches (422 for `POST /api/stock/add`, `status="bag_signature_mismatch"` for bulk-import rows) — back-compat when absent
- Tighten `find_by_bag_signature` guard from `isalnum()` to the same `[a-f0-9]{64}` regex
- Update bulk-import summary dict to include `bag_signature_mismatch` counter

## Key implementation note
Python's `str.strip()` strips `\x1c–\x1f` (FS/GS/RS/US field-separator control chars), but JavaScript's `.trim()` does **not**. A JS-compatible trim function is used so bag codes ending with a field separator hash identically in both implementations (the fixture `FOO-1 with trailing GS` was the failing case).

## Tests
- `backend/tests/test_bag_signature_parity.py` — fixture-driven parity test against `web/src/lib/__fixtures__/bagSignatures.json`
- `backend/tests/test_bag_signature_validation.py` — schema constraint + server-side mismatch tests for both write paths + `find_by_bag_signature` guard
- Backend: **490 passed, 2 skipped, 3 xfailed**
- Frontend build: **passed**

### DIFF
--- a/backend/app/api/routes/parts.py
+++ b/backend/app/api/routes/parts.py
@@ -1,9 +1,9 @@
 from __future__ import annotations
 
+import os
+import re
 from datetime import datetime, timezone
 from uuid import UUID
-
-import os
 
 from fastapi import APIRouter, Depends, HTTPException, Query, Request, status
 from fastapi.responses import FileResponse
@@ -383,13 +383,12 @@ def bulk_delete_parts(request: Request, payload: BulkDeleteIn, db: DbSession, ws
     )
 
 
-_BAG_SIG_RE_STR = r"^[a-f0-9]{64}$"
+_BAG_SIG_RE = re.compile(r"[a-f0-9]{64}")
 
 
 def _is_valid_bag_signature(s: str) -> bool:
     """Return True iff ``s`` is a 64-char lowercase hex string (SHA-256 digest)."""
-    import re
-    return bool(re.fullmatch(r"[a-f0-9]{64}", s))
+    return bool(_BAG_SIG_RE.fullmatch(s))
 
 
 @router.get("/by-bag-signature/{signature}")
--- a/backend/app/api/routes/parts.py
+++ b/backend/app/api/routes/parts.py
@@ -34,6 +34,7 @@ from app.domain.parts.schemas import (
     ScanImportRow,
     SubstituteIn,
 )
+from app.domain.parts.services.bag_signature import compute_bag_signature
 from app.domain.parts.services.assets import fetch_provider_asset
 from app.domain.stock.models import StockEntry
 from app.domain.stock.schemas import AddStockIn, LotInput
@@ -382,6 +383,15 @@ def bulk_delete_parts(request: Request, payload: BulkDeleteIn, db: DbSession, ws
     )
 
 
+_BAG_SIG_RE_STR = r"^[a-f0-9]{64}$"
+
+
+def _is_valid_bag_signature(s: str) -> bool:
+    """Return True iff ``s`` is a 64-char lowercase hex string (SHA-256 digest)."""
+    import re
+    return bool(re.fullmatch(r"[a-f0-9]{64}", s))
+
+
 @router.get("/by-bag-signature/{signature}")
 def find_by_bag_signature(signature: str, db: DbSession, ws: CurrentWorkspace):
     """Look up the most recent stock_entry whose bag_signature matches.
@@ -389,9 +399,10 @@ def find_by_bag_signature(signature: str, db: DbSession, ws: CurrentWorkspace):
     bag again, the frontend hits this endpoint to surface the prior
     import inline instead of waiting for the bulk-import POST to come
     back with a `bag_rescan` status."""
-    if len(signature) != 64 or not signature.isalnum():
-        # SHA-256 hex digest is exactly 64 hex chars. Reject obvious junk
-        # so this can't be abused for prefix-scan probing.
+    if not _is_valid_bag_signature(signature):
+        # SHA-256 hex digest is exactly 64 lower-case hex chars.  Reject
+        # anything that doesn't match (old code accepted upper-case via
+        # .isalnum(); tightened by BE2-015).
         return ok(None)
     prior = db.execute(
         select(StockEntry)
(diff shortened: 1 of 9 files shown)


matescb:

(Author == reviewer; using `gh pr comment` per self-review fallback.)

**Findings (1):**

1. **[high]** `backend/app/domain/parts/services/bag_signature.py:122` — Python pipeline does **two** `_js_trim` calls (before AND after `_normalise_control_pictures`); the TS `bagSignature()` in `web/src/lib/bagCode.ts:178` only trims once, *before* `normalizeControlPictures`. Result: any raw bag whose normalised form ends or begins in JS-whitespace after picture replacement (most relevantly `␠` → SP) hashes differently on the two sides. Concrete divergence (Python current vs. TS-faithful, verified locally):

   - `'FOO␠'` → py `9520437c…0fd3`, ts `034ff432…95c3`
   - `'␠FOO'` → py `9520437c…0fd3`, ts `e44e4a8e…d078`
   - `'FOO␠␠'` → py `9520437c…0fd3`, ts `651661fb…4e55`

   The fixture `web/src/lib/__fixtures__/bagSignatures.json` doesn't probe leading/trailing `␠` so the parity test passes, but this violates the load-bearing invariant in `CLAUDE.md` (`bag_signature` normalisation order must match exactly — signature is the only stable correlation key). The verification path is opt-in today (no FE caller currently sends `raw_bag_code`), so it's dormant — but the moment a FE caller is added, real bags whose normalised tail is `␠` would be falsely 422-rejected (`POST /api/stock/add`) or returned as `bag_signature_mismatch` in bulk import.

   **Fix:** drop the second `_js_trim` call in `compute_bag_signature` (and the docstring step #3 that justifies it) — TS intentionally trims only once. Then add a fixture entry with leading/trailing `␠` so the parity test catches the regression.

**Nits (low, non-blocking):**

- `backend/app/api/routes/parts.py:386` — `_BAG_SIG_RE_STR` defined but unused (the regex is inlined in `_is_valid_bag_signature`).
- `backend/app/api/routes/parts.py:392` — `import re` inside the function instead of at module top.

matescb:
Addressed the high-severity finding and both nits.

**High — second JS-trim removed.** `compute_bag_signature` now mirrors `web/src/lib/bagCode.ts::bagSignature` exactly: one `_js_trim`, then `_normalise_control_pictures`, then SHA-256. The post-trim no longer runs, so `'FOO␠'`, `'␠FOO'`, `'FOO␠␠'` (and any other bag whose normalised tail/head is an ASCII space exposed by `␠` → SP) hash identically on FE and BE. Updated docstring, dropped step #3 that justified the redundant trim, and added an inline `NB:` comment so this isn't reverted on autopilot.

**Fixture extended.** Added three entries to `web/src/lib/__fixtures__/bagSignatures.json` probing leading/trailing `␠`. Both the FE parity test (`bagCode.test.ts` → `bagSignature fixture parity`) and the BE parity test (`tests/test_bag_signature_parity.py::test_compute_bag_signature_parity`) now exercise the regression. Verified locally that the new digests match the TS-faithful pipeline (and would have failed on the old double-trim path).

**Nits.** `import re` hoisted to module top in `backend/app/api/routes/parts.py`; the dead `_BAG_SIG_RE_STR` string was replaced with a compiled `_BAG_SIG_RE = re.compile(r"[a-f0-9]{64}")` that `_is_valid_bag_signature` now uses directly.

Also merged latest `origin/main` (auto-merge, no conflicts).

Commit: 432e82d

matescb:
Rebased onto origin/main and fixed conftest.py to use `alembic upgrade heads` (plural) — origin/main now has two Alembic heads (0025 and 0029). This should fix the backend-tests CI failure.

matescb:

Verified parity: `compute_bag_signature` now mirrors TS `bagSignature` exactly (single `_js_trim` before `_normalise_control_pictures`, then SHA-256). Reproduced the three new fixture digests (`'FOO␠'`, `'␠FOO'`, `'FOO␠␠'`) independently against the TS-faithful pipeline — all match. Closing tracking issue #214.

PR #207, opened 2026-05-02
fix(fe): surface PartCreate refresh-provider failure + reset stale 409 state (#49)

### LINKED ISSUE #49: [FE2-011] PartCreate "lookup → create → refresh-from-provider" does three sequential round-trips with no rollback
**Source:** [`docs/teardown/frontend.md`](https://github.com/matescb/stockManager/blob/main/docs/teardown/frontend.md) (v2 teardown) — search for `### FE2-011:`

**Area:** frontend  •  **Priority:** medium

---

Severity: **Medium**

Evidence:
- `web/src/routes/parts/PartCreate.tsx:67-83` — flow: `api.post("/parts", ...)` → `api.post("/parts/{id}/refresh-from-provider")`. The refresh is wrapped in a swallow-all `try/catch`; the comment says "non-fatal" but failures are also silent (no toast, no per-part flag).
- The 409-on-MPN handler (`PartCreate.tsx:84-92`) returns early but does not clear `hasLookup`/`specs`/`imageUrl` state, so retrying with a new MPN re-applies the previous lookup's specs to the wrong part.
- If `refresh-from-provider` 5xx's after the part is created, the user lands on `PartInfo` with a part missing all provider-side data and no UI hint that a refresh is needed.

Impact:

Half-created parts show up in the library with the correct identity but no specs, image, or datasheet — and the user doesn't know it's their job to click Refresh. State leaks across MPN attempts cause the lookup preview to lie about a different MPN.

Fix instruction:

On 409, also reset `setDatasheetUrl(null)`, `setImageUrl(null)`, `setSpecs([])`, `setHasLookup(false)`. After the create+refresh combo, surface an inline "Provider refresh failed — retry?" banner if the second call rejected, instead of silently swallowing. Better: a single backend `POST /parts/with-lookup` that does both atomically.

---

matescb:
Closes #49

## Summary
- Reset hasLookup/specs/imageUrl/datasheetUrl on 409 so next MPN entry shows clean preview
- Replace swallow-all catch on refresh-from-provider with refreshFailed state + inline banner
- Banner offers Retry refresh (re-POSTs, navigates on success) and Open part anyway (immediate nav)
- Create button disabled while banner is shown

## Tests
Backend: N/A
Frontend build: passed

### DIFF
--- a/web/src/routes/parts/PartCreate.tsx
+++ b/web/src/routes/parts/PartCreate.tsx
@@ -54,6 +54,10 @@ export default function PartCreate() {
   const [imageUrl, setImageUrl] = useState<string | null>(null);
   const [specs, setSpecs] = useState<ProviderSpec[]>([]);
   const [hasLookup, setHasLookup] = useState(false);
+  // When refresh-from-provider fails after a successful create, we surface
+  // an inline banner instead of silently swallowing the error. The part
+  // is valid — just missing provider-side fields — so we never DELETE it.
+  const [refreshFailed, setRefreshFailed] = useState<{ partId: string } | null>(null);
   const { data: storage } = useQuery({ queryKey: useWsKey("storage"), queryFn: () => api.get<StorageLocation[]>("/storage") });
 
   // FE2-006: gate concurrent submits via mutationKey so a double-click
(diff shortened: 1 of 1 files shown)


matescb:

PR #211, opened 2026-05-02
fix(api): request_id in all error responses + validation failure logging (#61)

### LINKED ISSUE #61: [BE2-012] `validation_exception_handler` does not log validation failures, and `http_exception_handler` does not include a request_id
**Source:** [`docs/teardown/backend.md`](https://github.com/matescb/stockManager/blob/main/docs/teardown/backend.md) (v2 teardown) — search for `### BE2-012:`

**Area:** backend  •  **Priority:** medium

---

Severity: **Medium**

Evidence:
- `backend/app/core/responses.py:24` `http_exception_handler` logs only on 5xx (`if exc.status_code >= 500`).
- `backend/app/core/responses.py:71` `validation_exception_handler` doesn't log at all.
- No `X-Request-ID` middleware anywhere; no request-id is included in the JSON envelope.
- `backend/app/main.py:14` configures structured logging but doesn't add a request-id `LogFilter`.

Impact:

When a user reports "I tried to create a part and got an error", the only correlation key is timestamp + endpoint. There's no propagated request id between logs, Sentry, and the JSON returned to the operator. 4xx responses (the 90% case for client errors) silently disappear from the log. Validation failures — which are a strong signal of a frontend regression — leave no trace.

Fix instruction:

Add a small middleware that mints `request.state.request_id = uuid.uuid4().hex` and emits it on every log line and as `X-Request-Id` response header. Spread `request_id` into the `err()` envelope. Log validation failures at INFO with the field list so a frontend regression is visible in the journal.

---

matescb:
Closes #61

## Summary
- Add `RequestIdMiddleware` that mints/reuses a hex request_id on every request, sets it on `request.state.request_id`, and echoes it as `X-Request-Id` response header
- Add `RequestIdFilter` to `core/logging.py`; attaches to root handler so every log record emitted during a request carries `request_id` automatically via ContextVar
- Extend `err()` with optional `request_id` kwarg; inject into error envelopes so the JSON body always carries the correlation key
- Update `http_exception_handler` to log all 4xx at INFO (was: only 5xx) and inject `request_id` into body
- Update `validation_exception_handler` to log at INFO with field list and inject `request_id`
- Rename `0021_invitation_lookup_index.py` → `0022` to resolve duplicate revision conflict introduced by two PRs both landing as `0021` on main
- Fix `alembic/env.py` `fileConfig` call to pass `disable_existing_loggers=False` so app loggers are not silently disabled during test migrations

## Tests
Backend: **496 passed**, 1 skipped, 3 deselected, 3 xfailed
- 10 new tests added: `test_request_id_middleware`, `test_validation_logging`, `test_4xx_logging`

Frontend build: N/A (backend-only change)

### DIFF
--- a/backend/alembic/env.py
+++ b/backend/alembic/env.py
@@ -14,7 +14,12 @@ config = context.config
 config.set_main_option("sqlalchemy.url", os.environ.get("DATABASE_URL", settings().DATABASE_URL))
 
 if config.config_file_name is not None:
-    fileConfig(config.config_file_name)
+    # disable_existing_loggers=False preserves app loggers that were
+    # already configured by configure_logging() — without this the alembic
+    # fileConfig() call silently sets logger.disabled=True on every app
+    # logger, causing http_exception_handler and validation_exception_handler
+    # to emit no records during tests (BE2-012 / issue #61).
+    fileConfig(config.config_file_name, disable_existing_loggers=False)
 
 target_metadata = Base.metadata
 
--- a/backend/app/core/logging.py
+++ b/backend/app/core/logging.py
@@ -31,6 +31,25 @@ from typing import Any
 _CONFIGURED = False
 
 
+class RequestIdFilter(logging.Filter):
+    """Inject the current request-id (from the contextvar set by
+    `RequestIdMiddleware`) into every log record while a request is
+    being handled.  Records emitted outside a request context get
+    ``request_id=None`` so downstream log shippers can filter on it."""
+
+    def filter(self, record: logging.LogRecord) -> bool:
+        # Local import to avoid a circular import at module load time
+        # (request_id.py → logging.py on configure_logging, but
+        # logging.py would need request_id.py only at call time).
+        try:
+            from app.core.request_id import request_id_var  # noqa: PLC0415
+
+            record.request_id = request_id_var.get()
+        except Exception:  # noqa: BLE001
+            record.request_id = None
+        return True
+
+
 class _JsonFormatter(logging.Formatter):
     """Compact JSON-per-line formatter. No multi-line stack traces in
     the body; tracebacks land in `traceback` as a single string."""
@@ -90,6 +109,7 @@ def configure_logging() -> None:
     # — rip them out and replace with ours so format is consistent.
     for h in list(root.handlers):
         root.removeHandler(h)
+    handler.addFilter(RequestIdFilter())
     root.addHandler(handler)
 
     # Quiet down noisy stdlib / lib loggers — these are too chatty at
--- a/backend/app/core/request_id.py
+++ b/backend/app/core/request_id.py
@@ -1,0 +1,75 @@
+"""Request-ID middleware and context-variable helpers.
+
+BE2-012 / issue #61. Every inbound request gets a `request_id` (hex UUID)
+minted and stored on `request.state.request_id`.  The same value is
+propagated via:
+
+- `X-Request-Id` response header (so the browser / curl caller can copy it)
+- a `ContextVar` that `LogFilter` (see `core/logging.py`) injects into every
+  log record emitted during that request's stack, so operators can grep the
+  journal for a single id.
+- `core/responses` error envelopes so the frontend can surface it on error
+  modals.
+
+If the caller sends an `X-Request-Id` header it is reused *if and only if*
+it matches the allowed shape (1–64 hexadecimal characters, case-insensitive).
+A malformed inbound value is silently replaced with a fresh one — never
+rejected, never logged as an error.
+"""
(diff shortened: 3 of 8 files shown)


matescb:
### claude-review — request changes (self-review fallback)

**Verdict:** request-changes (highest severity: medium)

**Findings (1):**

1. **[medium]** `backend/app/main.py:318-319` — `RequestIdMiddleware` is registered *before* `CsrfOriginMiddleware` and `CORSMiddleware`. Starlette stacks middleware LIFO: the last `add_middleware()` call becomes the **outermost** wrapper. The current order makes `RequestIdMiddleware` the **innermost** middleware, so when `CsrfOriginMiddleware` short-circuits with a 403 (`main.py:272-281`) or `CORSMiddleware` handles a preflight, the request never reaches `RequestIdMiddleware`: no `X-Request-Id` response header, no `request_id` in body, no contextvar set, no Sentry tag. The new docstring on the middleware and the inline comment in `main.py` both state "must run BEFORE CORS / CSRF so request.state.request_id is set even when those middlewares short-circuit" — but the placement achieves the opposite. Suggest: move `app.add_middleware(RequestIdMiddleware)` to **after** the CORS `add_middleware` call (so it wraps everything else), and enrich the CSRF rejection JSONResponse in `CsrfOriginMiddleware.dispatch` to surface the now-available `request.state.request_id`. Tests don't catch this because every test hits either GET `/api/health` (CSRF passes through GETs) or the CSRF-exempt `/api/auth/login` — add a POST to a CSRF-protected path with a bad Origin to lock the invariant down.

Other items reviewed and OK:
- API envelope intact: `request_id` is added as a top-level sibling alongside `data`/`status`/`errors`; the `HTTPException(detail=…)` dict-spread is preserved and no existing detail dict uses the `request_id` key.
- Validation handler logs only `path`, `fields` (loc + msg), and `request_id` — no request body capture, so passwords/tokens can't leak.
- Minted ids use `uuid.uuid4().hex` (122 bits). Inbound reuse is bounded to `^[0-9a-fA-F]{1,64}$` and is correlation-only (not auth) so caller-supplied predictable ids are acceptable.
- `alembic/env.py` `disable_existing_loggers=False` is the right fix for tests dropping app loggers.

matescb:
Addressed the medium-severity finding from claude-review (commit 0f409363714ad704dbe6a70e48dceefa09b1f042):

- Moved `app.add_middleware(RequestIdMiddleware)` to be the **last** add_middleware() call so it becomes the outermost wrapper. With Starlette's LIFO stacking, this means RequestId now wraps CORS and CSRF — short-circuit responses (CSRF 403, CORS preflight) still see `request.state.request_id`, get it logged via the contextvar filter, and carry the `X-Request-Id` header.
- Enriched `CsrfOriginMiddleware`'s 403 `JSONResponse` to surface the now-available `request.state.request_id` on both the body (top-level `request_id`) and the `X-Request-Id` response header.
- Fixed the misleading docstring on `RequestIdMiddleware` ("before CORS/CSRF" → "wrap CORS/CSRF / registered LAST").
- Added `test_csrf_rejection_carries_request_id` that POSTs to a CSRF-protected path with a hostile Origin, asserting the 403 carries the inbound `X-Request-Id` verbatim in both header and body. This locks down the LIFO ordering invariant — regressing the add_middleware() order will break the assertion.

Targeted pytest: 17/17 passed (test_request_id_middleware, test_validation_logging, test_4xx_logging, plus all CSRF / origin / auth_login tests).

matescb:
Re-review at `0f40936`: middleware ordering fix verified, merging.

**Fixed**: `backend/app/main.py` adds middleware in correct LIFO order — `CsrfOriginMiddleware` (line 338, innermost) → `CORSMiddleware` (344) → `RequestIdMiddleware` (353, outermost). CSRF and CORS short-circuits now flow through the request-id middleware and get `X-Request-Id` header + `request_id` body key. The new `test_csrf_rejection_carries_request_id` test pins the invariant — a regression in add_middleware() order would break it loudly.

**Other checks pass:**
- API envelope `{ data, status }` preserved; `request_id` added as top-level sibling key.
- Validation handler logs only field names + pydantic messages + path — no body, no PII, no password/token capture.
- `request_id` is opaque (`uuid.uuid4().hex`, 128-bit random), not sequential.
- Inbound `X-Request-Id` validated against `^[0-9a-fA-F]{1,64}$` and silently replaced if malformed (prevents log/header injection; tested with `../../../etc/passwd` and 65-char overflow).
- CSRF rejection body defensively guards `if rid:`.
- Sentry tag set inside try/except.
- `disable_existing_loggers=False` in `alembic/env.py` is the right fix for app loggers being disabled during tests.

PR #223, opened 2026-05-02
fix(stock): enforce storage constraints via current stock (#35)

### LINKED ISSUE #35: [BE-004] Storage Constraints Are Incomplete And Historically Wrong
**Source:** [`docs/claude-review-issues.md`](https://github.com/matescb/stockManager/blob/main/docs/claude-review-issues.md) (v1 review) — search for `### BE-004:`

**Area:** backend  •  **Priority:** high

---

Severity: **High**

Evidence:

- `StorageLocation` exposes `single_part_only` and `existing_parts_only`.
- `add_stock()` does not enforce either flag.
- `move_stock()` enforces `single_part_only` by counting any historical row for another part, not current positive stock.

Impact:

Add-stock can mix parts into restricted storage. Move-stock can falsely reject after another part was fully removed. `existing_parts_only` appears unused.

Fix instruction:

Centralize destination-storage validation for add, move, receive, and build output. Enforce against grouped current on-hand quantity, not historical row count. Define and test `existing_parts_only`.

---

matescb:
Closes #35

## Summary
- Add `_enforce_storage_constraints()` private helper in `stock/service.py` that uses `stock_for_storage()` (current positive balances) instead of historical row counts
- `add_stock` now enforces both `single_part_only` and `existing_parts_only` when a storage location is provided
- `move_stock` replaces the old broken `SELECT COUNT(*)` historical check with the correct helper
- `StockConflictError(StockError)` subclass added with `constraint` + `storage_location_id` attributes
- Routes catch `StockConflictError` → 409 with `{"message", "constraint", "storage_location_id"}` in detail
- Fix `conftest.py` to use alembic `"heads"` (plural) to handle the existing multi-head migration chain

## Tests
9 new tests in `backend/tests/test_storage_constraints.py`:
1. `single_part_only` rejects add of part B when part A has positive stock → 409
2. `single_part_only` allows add of same part A (re-adding)
3. `single_part_only` allows add of part B after all part-A stock removed (regression fix)
4. `single_part_only` rejects move of part B into location holding part A
5. `single_part_only` allows move of part B to historically-held but now-empty location (regression fix)
6. `existing_parts_only` rejects add of part with no prior history → 409
7. `existing_parts_only` allows add of part with prior positive history (even if currently zero)
8. Combined flags behavior
9. Workspace isolation via ws-filtered `stock_for_storage`

Backend: 546 passed, 1 skipped, 3 xfailed
Frontend: N/A

### DIFF
--- a/backend/app/api/routes/stock.py
+++ b/backend/app/api/routes/stock.py
@@ -11,6 +11,7 @@ from app.domain.stock.schemas import (
     RemoveStockIn,
 )
 from app.domain.stock.service import (
+    StockConflictError,
     StockError,
     add_stock,
     adjust_stock,
@@ -50,6 +51,15 @@ def add(
 ) -> Envelope[dict]:
     try:
         e = add_stock(db, workspace_id=ws.id, user_id=user.id, payload=payload)
+    except StockConflictError as exc:
+        raise HTTPException(
+            status_code=status.HTTP_409_CONFLICT,
+            detail={
+                "message": str(exc),
+                "constraint": exc.constraint,
+                "storage_location_id": str(exc.storage_location_id),
+            },
+        )
     except StockError as exc:
         raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=str(exc))
     return ok(_serialize_entry(e))
@@ -70,6 +80,15 @@ def remove(
 def move(payload: MoveStockIn, db: DbSession, ws: CurrentWorkspace, user: CurrentUser):
     try:
         out_e, in_e = move_stock(db, workspace_id=ws.id, user_id=user.id, payload=payload)
+    except StockConflictError as exc:
+        raise HTTPException(
+            status_code=status.HTTP_409_CONFLICT,
+            detail={
+                "message": str(exc),
+                "constraint": exc.constraint,
+                "storage_location_id": str(exc.storage_location_id),
+            },
+        )
     except StockError as exc:
         raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=str(exc))
     return ok({"out": _serialize_entry(out_e), "in": _serialize_entry(in_e)})
--- a/backend/app/domain/stock/service.py
+++ b/backend/app/domain/stock/service.py
@@ -104,6 +104,41 @@ def lock_parts_for_stock_write(
         _lock_for_stock_write(db, workspace_id=workspace_id, part_id=pid)
 
 
+def _lock_for_storage_constraint(
+    db: Session, *, workspace_id: UUID, storage_location_id: UUID
+) -> None:
+    """Per-(workspace, storage) advisory lock for storage-constraint reads.
+
+    The per-(workspace, part) lock from ``_lock_for_stock_write`` only
+    serialises writers that touch the *same* part. ``single_part_only``
+    and ``existing_parts_only`` are cross-part invariants on the
+    *destination location*: two concurrent ``add_stock`` / ``move_stock``
+    calls for **different parts** targeting the same single_part_only
+    bin each hold a different per-part lock, both read
+    ``stock_for_storage(loc) == []``, both pass the check, and both
+    insert their StockEntry — leaving the bin holding two parts.
+
+    This second lock closes the cross-part race by serialising every
+    write whose destination is the same location. There is no DB-level
+    CHECK or trigger backing single_part_only / existing_parts_only
+    (the 0013 trigger only enforces non-negative balances), so the
+    application-level lock is the only line of defence.
+
+    **Lock ordering**: callers always acquire the per-part lock first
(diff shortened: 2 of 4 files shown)


matescb:

**Findings (2):**

1. **[high]** `backend/app/domain/stock/service.py:296-308` (and the `move_stock` callsite around line 540) — The advisory lock used for storage-constraint enforcement is keyed on `(workspace_id, part_id)`, so two concurrent `add_stock` / `move_stock` calls for **different parts** targeting the same `single_part_only` storage location each hold a *different* lock and run in parallel. Both can read `stock_for_storage(loc) = []`, both pass the single-part-only check, and both insert their `StockEntry` — leaving the location holding two different parts in violation of the invariant. There is no DB-level CHECK or trigger backing this constraint (the 0013 trigger only enforces non-negative balances), so the application-level lock is the only line of defence and it does not cover this race. The new helper's docstring claims "Must be called inside the advisory lock for (workspace_id, part_id) so the read-check-write sequence is serialised" — true for same-part races, false for cross-part races against the same destination. The same hole exists for `existing_parts_only` (a concurrent removal of the prior positive entry can race with the prior-history check, though this is more theoretical since prior history is monotonic per part). Suggest: also take a per-storage advisory lock when a storage_location is involved (e.g. `pg_advisory_xact_lock(hashtextextended(f"{workspace_id}:storage:{storage.id}", 0))`) before the constraint read; ordering must be deterministic with the existing per-part lock to avoid deadlock (e.g. always acquire storage lock after the part lock, or wrap both in a sorted helper). Add a concurrency regression test that fires two concurrent `add_stock` calls for parts A and B at the same `single_part_only` empty location and asserts exactly one succeeds.

2. **[low]** `backend/tests/test_storage_constraints.py:217-232` — `test_existing_parts_only_allows_part_with_prior_positive_history` first creates `loc` + `staging`, adds stock to staging, and PATCHes `loc` — then abandons all of that and starts over with `loc2` / `part2`. The first block has no assertion path that exercises `loc` and is effectively dead code. Drop it; the `loc2` half is the real test.

CI is green and the PR fixes a real correctness bug (the old historical-row check was wrong) — but the new check is still not race-safe, which the codebase guidance (storage constraints must be enforced atomically with the insert) calls out as a high-severity issue. Please address #1 before merge.

 review complete at 2026-05-02T14:00:00Z

matescb:
Addressed both review findings in 424c511:

**Finding 1 (high) — cross-part race against single_part_only / existing_parts_only**
Added `_lock_for_storage_constraint` (`pg_advisory_xact_lock` keyed on `f"{workspace_id}:storage:{storage_id}"`) and acquire it inside `_enforce_storage_constraints` before the read. Lock ordering is fixed: per-part lock (`_lock_for_stock_write`) first, per-storage lock second — no AB/BA deadlock with `lock_parts_for_stock_write`. Two concurrent `add_stock` / `move_stock` calls for different parts targeting the same single_part_only destination now serialise on the storage lock so the second sees the first's StockEntry and 409s.

Added a threaded regression test `test_single_part_only_cross_part_race_blocked` that fires concurrent `POST /api/stock/add` for parts A and B at the same empty single_part_only location and asserts exactly one succeeds + one returns 409.

**Finding 2 (low) — dead loc/staging block in `test_existing_parts_only_allows_part_with_prior_positive_history`**
Removed; the test now keeps only the meaningful path (create unflagged loc → seed prior positive history → PATCH flag on → re-add succeeds).

matescb:
**Approve.** Lock scope is now correct.

- `_lock_for_storage_constraint` keys on `f"{workspace_id}:storage:{storage_location_id}"`, so two concurrent writers targeting the same `single_part_only` bin with **different** parts now serialise on the destination, not their own per-part locks. The cross-part TOCTOU from the prior review is closed.
- Lock ordering is fixed (part → storage) and only ever in that direction, so no AB/BA cycle.
- `single_part_only` check now uses `stock_for_storage()` (canonical current-balance service per CLAUDE.md), not a historical `SELECT COUNT(*)`. Regression tests cover the "previously held A, now empty → B accepted" case for both add and move.
- `move_stock` no longer carries its own historical-count branch; both `add_stock` and `move_stock` route through the helper. Single source of truth.
- Concurrency test `test_single_part_only_cross_part_race_blocked` exercises the exact race the previous SHA failed (two threads, two parts, same SPO bin, barrier-synchronised) and asserts 1 success + 1 structured 409.
- Workspace isolation pinned by `test_workspace_isolation_storage_constraints`.
- 409 detail surfaces `constraint` + `storage_location_id` — clean envelope.

Out of scope but worth flagging for a follow-up (not blocking): `orders.receive` / `builds.consume` / `builds.apply_reservations` write `StockEntry` rows directly without going through `_enforce_storage_constraints`, so a receive/build can still drop a part into a `single_part_only` or `existing_parts_only` bin. The PR docstring acknowledges this. Existing gap, not introduced here.

Merging.

PR #241, opened 2026-05-02
fix: restore linear Alembic numbering after migration race

matescb:
## Summary
- restore bulk-import idempotency as revision 0034 after 0033
- keep workspace catalog tokens as revision 0035 after 0034
- fix the workspace migration docstring metadata to match its revision

## Verification
- script-based migration graph check reports `heads: ['0035']`, no duplicate revisions, and no missing parents

### DIFF
--- a/backend/alembic/versions/0034_bulk_import_idempotency.py
+++ b/backend/alembic/versions/0034_bulk_import_idempotency.py
@@ -1,12 +1,12 @@
 """Add bulk_import_idempotency table (BE2-003).
 
-Revision ID: 0035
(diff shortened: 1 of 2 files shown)


matescb:
Merge/update note: branch rebased onto current main after #390. No remaining merge-order dependency observed; conflicts resolved by rebase only. Local checks: uv run alembic heads => 0039 (head); uv run ruff check on 0034/0035 migrations; uv run python -m pytest -q tests/test_migrations.py -m slow; git diff --check.

PR #260, opened 2026-05-02
fix(orders): acquire stock lock before reading entries to eliminate receive race (#247)

### LINKED ISSUE #247: [TEST-003 follow-up] Order receive race not actually fixed; tests are xfail

PR #131 added concurrency tests for stock-remove and builds-consume that pass cleanly. The orders-receive tests in the same PR are `@pytest.mark.xfail(strict=False)` because they uncovered a real bug that was never fixed.

## Evidence

`backend/app/domain/orders/service.py:53` reads `OrderEntry` rows BEFORE acquiring the advisory lock:

```python
entries_by_id: dict[UUID, OrderEntry] = {
    e.id: e
    for e in db.query(OrderEntry)
    .filter(OrderEntry.workspace_id == workspace_id, OrderEntry.order_id == order.id)
    .all()
}
# Lock acquired AFTER read:
lock_parts_for_stock_write(db, workspace_id=workspace_id, part_ids=[...])
```

Then uses stale `oe.quantity_received` at `:84` for the `outstanding` check. Two concurrent receives both see the same outstanding, both pass, both write — over-receive.

`backend/tests/test_orders_receive_concurrency.py:70` admits it:

```python
@pytest.mark.xfail(strict=False, reason="Real bug surfaced by this regression test (BE-001 follow-up). Fix: re-load entries (or SELECT … FOR UPDATE on the OrderEntry rows) AFTER the advisory lock.")
```

`strict=False` means the test passes whether the bug fires or not — it's documentation, not a regression net.

## Suggested fix

Either:

**(a)** re-load entries after the lock:
```python
lock_parts_for_stock_write(db, workspace_id=workspace_id, part_ids=[...])
db.expire_all()  # or refresh per row
entries_by_id = { e.id: e for e in db.query(OrderEntry).filter(...).all() }
```

**(b)** `SELECT ... FOR UPDATE` on the rows the receive will touch:
```python
entries_by_id = {
    e.id: e for e in db.query(OrderEntry)
    .filter(OrderEntry.id.in_([line.order_entry_id for line in payload.lines]))
    .with_for_update()
    .all()
}
```

(b) is more targeted and avoids the read-the-world reload.

After the fix, flip the test marker from `xfail(strict=False)` to a plain assertion.

## Verify

The existing test in `test_orders_receive_concurrency.py` becomes the regression net once xfail is removed.

matescb:
Closes #247

## Summary
- Restructure `receive` in `domain/orders/service.py` to acquire `lock_parts_for_stock_write` BEFORE reading `OrderEntry` rows
- Add a lightweight preliminary query to collect `part_ids` for the lock call (preserves workspace isolation — still filters by `workspace_id`)
- Re-query `OrderEntry` rows with `.with_for_update()` after lock acquisition so `quantity_received` is authoritative and any concurrent commits are visible
- Remove `@pytest.mark.xfail` from `test_concurrent_receive_cannot_overshoot_qty_ordered` and `test_partial_receive_serialises` in `tests/test_orders_receive_concurrency.py`

## Tests
Backend: 684 passed, 2 skipped, 3 deselected, 1 xfailed (pre-existing, unrelated), 63 warnings
Frontend: N/A

### DIFF
--- a/backend/app/domain/orders/service.py
+++ b/backend/app/domain/orders/service.py
@@ -50,23 +50,35 @@ def receive(
     if order.status == "cancelled":
         raise OrderError("order is cancelled")
 
+    # BE2-001 / #247: acquire the advisory stock-write lock BEFORE reading
+    # OrderEntry rows. Reading first and locking second is a TOCTOU race —
+    # a concurrent receive can slip through the `outstanding` guard while
+    # both threads hold stale in-memory `quantity_received` values.
+    #
+    # Fix: run a lightweight preliminary query to collect the part_ids for
+    # the lock call, acquire the lock, then re-query entries with
+    # FOR UPDATE so the values we act on are authoritative and pinned
+    # for the rest of this transaction.
+    preliminary_entries = (
+        db.query(OrderEntry.id, OrderEntry.part_id)
+        .filter(OrderEntry.workspace_id == workspace_id, OrderEntry.order_id == order.id)
+        .all()
+    )
+    lock_parts_for_stock_write(
+        db,
+        workspace_id=workspace_id,
+        part_ids=[row.part_id for row in preliminary_entries if row.part_id is not None],
+    )
+
+    # Re-query with FOR UPDATE after the lock so quantity_received reflects
+    # any in-flight writes that committed before we acquired the lock.
     entries_by_id: dict[UUID, OrderEntry] = {
         e.id: e
         for e in db.query(OrderEntry)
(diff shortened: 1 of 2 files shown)


matescb:
Approved by claude-review (no medium+ findings, CI green). Verified: lock-ordering unchanged (sorted-UUID inside `lock_parts_for_stock_write`), per-(ws, part) advisory key consistent with builds.consume / apply_reservations / release_reservations callers, preliminary query preserves `workspace_id` filter, re-query under `with_for_update()` is the right TOCTOU fix, and both un-xfailed regression tests exercise the race.

PR #263, opened 2026-05-02
fix(assets): validate provider asset magic bytes before writing (#246)

matescb:
Closes #246

## Summary
- Add `_MAGIC` dict mapping extension to byte prefix signatures (jpg, png, pdf, gif, webp)
- Add `_sniff_ext(header: bytes) -> str | None` helper that matches leading bytes against `_MAGIC`, with RIFF/WebP disambiguation
- After downloading provider asset body, call `_sniff_ext(body[:16])` and reject (return `None`) if the sniffed type contradicts the Content-Type-derived extension; unknown magic (sniff=None) passes through to preserve fail-tolerant behaviour
- Add 7 new unit tests in `test_provider_assets.py` covering mismatch rejection (PDF-as-PNG, JPEG-as-PDF, PNG-as-GIF), valid acceptance (matching magic), unknown-magic pass-through, and `_sniff_ext` helper in isolation

## Tests
Backend: 677 passed, 27/27 asset tests passed. Pre-existing failures in unrelated test files (test_integer_quantities, test_activity_pagination, test_history_limits, test_invitations) unchanged by this PR.
Frontend: N/A

PR #201, opened 2026-05-02
fix(fe): AttachmentsPanel client-side type/size guard (#52)

matescb:
Closes #52

## Summary
- Add `MAX_BYTES` (50 MB) constant and `ALLOWED_MIME_FOR_TYPE` map keyed on all six `file_type` values; `other` remains fully permissive on extension, still subject to size cap
- Validate in `doUpload()` before `api.upload()`, in `onDrop()`, and in the `onChange` handler — all short-circuit with `toast.error(...)` on failure, naming the allowed extensions or the size cap
- Wire `accept` attribute on `<input type="file">` computed via `useMemo` keyed off `fileType` so the OS picker pre-filters
- Update helper text near the dropzone to surface allowed types and size cap for the selected type
- Export `validateFile`, `humanSize`, `MAX_BYTES`, and `ALLOWED_MIME_FOR_TYPE` to enable pure unit testing without jsdom

## Tests
Backend: N/A
Frontend build: passed (95 tests pass, 30 new for AttachmentsPanel)

PR #290, opened 2026-05-03
fix(security): strict magic-byte validation — reject unknown sniffs (#246)

### LINKED ISSUE #246: [SEC-003 follow-up] Provider asset extension by MIME map, not magic-byte validated

Asset hardening (PR #32 / SEC2-006) added host allow-list, SSRF defenses, SVG rejection, and `nosniff` on serve. But the original SEC-003 specifically asked for **magic-byte validation of PNG/JPEG/WebP/PDF** — that part wasn't done.

## Evidence

`backend/app/domain/parts/services/assets.py:85`:

```python
def _ext_from_response(resp: httpx.Response, url: str) -> str:
    ct = (resp.headers.get("content-type") or "").split(";", 1)[0].strip().lower()
    if ct == "image/svg+xml":
        return "bin"
    if ct in _EXT_BY_MIME:
        return _EXT_BY_MIME[ct]
    by_url = _ext_from_url(url)
    ...
```

Trusts upstream `Content-Type` header. A malicious upstream can declare `image/png` and serve HTML.

## Suggested fix

Add magic-byte sniff (use `puremagic`, or hand-roll the 8-byte signatures — only 5 formats):

```python
_MAGIC = {
    b"\x89PNG\r\n\x1a\n": "png",
    b"\xff\xd8\xff": "jpg",
    b"RIFF": "webp",  # plus 'WEBP' at offset 8
    b"%PDF-": "pdf",
    b"GIF87a": "gif", b"GIF89a": "gif",
}
```

Validate the first N bytes of `resp.content` matches the declared `Content-Type`. Reject mismatches as `bin`.

**Keep GIF in the allow-set** (per operator decision — some providers serve GIF datasheets/images).

## Verify

Add `tests/test_provider_assets.py` cases:
- PNG body declared as PNG → accepted
- HTML body declared as `image/png` → rejected (lands as `bin`)
- GIF body declared as GIF → accepted
- Truncated body (<8 bytes) → rejected

matescb:
## Summary

PR #263 added magic-byte sniffing in `fetch_provider_asset`, but the check only rejected on a *known* sniff mismatch. When `_sniff_ext` returned `None` (unknown leading bytes), the body was silently accepted under whatever extension the upstream `Content-Type` declared — including `.png` / `.jpg` / `.webp` / `.pdf` / `.gif`. This is a content-type-confusion bypass: arbitrary attacker-controlled bytes could land on disk with an image/PDF extension and an image/PDF MIME on serve.

## The fix

`backend/app/domain/parts/services/assets.py` — tighten the gate so unknown sniffs are rejected:

```python
if ext != "bin":
    sniffed = _sniff_ext(body[:16])
    if sniffed != ext:
        log.warning(
            "provider asset rejected: magic bytes (%s) do not match "
            "declared extension (%s) from %s",
            sniffed or "<unknown>", ext, url,
        )
        return None
```

The `.bin` fallback path is unchanged (already covered by `Content-Disposition: attachment` + `application/octet-stream` in the serve route).

## Test changes

- Inverted `test_magic_bytes_unknown_body_with_known_ext_accepted` → `…_rejected`; also asserts no file is written.
- Added parametrised reject-tests covering `png` / `jpg` / `webp` / `pdf` / `gif`. GIF stays in the allow-list per the operator decision recorded on the issue; this case just pins the unknown-magic reject path for it.
- Updated `_resp()` defaults in both `test_provider_assets.py` and `test_assets.py` to use a real PNG-magic prefix so existing accept-path tests keep passing under the stricter check.

Test count: backend went from 729 → 731 (added 2 new test functions; the parametrised case expands to 5 IDs but pytest counts the test function once at collection summary).

```
731 passed, 2 skipped, 3 deselected, 63 warnings in 214.76s
```

## Hard invariants preserved

- Content-addressed storage path `{UPLOAD_DIR}/parts/{ws_id}/{sha}.{ext}` is unchanged — only the decision-to-write is gated.
- No change to the serve route, the URL structure, or the API envelope.

## Test plan

- [x] `pytest tests/test_provider_assets.py tests/test_assets.py -v` — 32 passed
- [x] Full backend suite — 731 passed, 2 skipped
- [x] Existing reject tests (HTML-as-PNG, mismatched-format, JPEG-as-PDF, PNG-as-GIF) still pass

Closes #246

### DIFF
--- a/backend/app/domain/parts/services/assets.py
+++ b/backend/app/domain/parts/services/assets.py
@@ -212,11 +212,11 @@ def fetch_provider_asset(url: str, workspace_id: str, kind: str) -> str | None:
     # those already carry a forced-download Content-Disposition when served.
     if ext != "bin":
         sniffed = _sniff_ext(body[:16])
-        if sniffed is not None and sniffed != ext:
+        if sniffed != ext:
             log.warning(
                 "provider asset rejected: magic bytes (%s) do not match "
                 "declared extension (%s) from %s",
-                sniffed,
+                sniffed or "<unknown>",
                 ext,
                 url,
             )
--- a/backend/tests/test_assets.py
+++ b/backend/tests/test_assets.py
@@ -22,7 +22,12 @@ def ws_id() -> str:
     return str(uuid.uuid4())
 
 
-def _resp(status_code: int = 200, body: bytes = b"image-bytes", content_type: str = "image/png") -> MagicMock:
+# Minimal PNG magic prefix — needed because magic-byte validation (#246)
+# now rejects bodies whose leading bytes don't match the declared ext.
+_PNG_BODY = b"\x89PNG\r\n\x1a\n" + b"image-bytes"
+
+
+def _resp(status_code: int = 200, body: bytes = _PNG_BODY, content_type: str = "image/png") -> MagicMock:
     r = MagicMock()
     r.status_code = status_code
     r.content = body
@@ -48,7 +53,7 @@ def test_fetch_writes_to_uploads_and_returns_local_path(monkeypatch, ws_id, tmp_
     sha = result.rsplit("/", 1)[-1].split(".")[0]
     on_disk = tmp_path / "parts" / ws_id / f"{sha}.png"
     assert on_disk.exists()
-    assert on_disk.read_bytes() == b"image-bytes"
+    assert on_disk.read_bytes() == _PNG_BODY
 
 
 def test_fetch_is_idempotent(monkeypatch, ws_id, tmp_path):
--- a/backend/tests/test_provider_assets.py
+++ b/backend/tests/test_provider_assets.py
@@ -26,7 +26,12 @@ from app.domain.parts.services import assets
 from app.main import app
 
 
-def _resp(status_code: int = 200, body: bytes = b"image-bytes", content_type: str = "image/png") -> MagicMock:
+# Default body carries valid PNG magic so the strict #246 magic-byte check
+# accepts it; tests that need other bodies override `body=` explicitly.
+_PNG_BODY = b"\x89PNG\r\n\x1a\n" + b"image-bytes"
+
+
+def _resp(status_code: int = 200, body: bytes = _PNG_BODY, content_type: str = "image/png") -> MagicMock:
     r = MagicMock()
     r.status_code = status_code
     r.content = body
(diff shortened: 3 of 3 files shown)


matescb:
**claude-review: APPROVE — strict magic-byte gate, correct, well-tested**

Reviewed against `.claude/prompts/pr-watcher.md` + `CLAUDE.md`.

**Fix correctness**
- One-character semantic tightening: `if sniffed is not None and sniffed != ext:` → `if sniffed != ext:`. Since the outer guard already excludes `ext == "bin"`, `ext` here is always one of `{jpg, png, pdf, gif, webp}`. `None != ext` is always True, so unknown sniffs now reject — closing the content-type-confusion bypass from #263.
- Log line updated to print `sniffed or "<unknown>"` so the warning still reads sensibly when `sniffed is None`.

**Critical gates (per watcher prompt)**
- Allow-list enumerated: PNG / JPEG / WebP / PDF / GIF — exactly the keys in `_MAGIC` and `_EXT_BY_MIME`. Matches #201/#263 alignment.
- Rejection returns `None`, not a generic 500. Caller falls back to the original upstream URL (no exception, graceful), and a warning is logged with sniffed/declared/url. FE behaviour unchanged.
- 4 MB / 10 MB caps untouched. `_MAX_BYTES = 10 * 1024 * 1024` (provider) and `MAX_UPLOAD_BYTES = 10 * 1024 * 1024` (user upload, separate path) both unchanged.
- Magic-byte check runs at `assets.py:214-224`, BEFORE the `os.makedirs`/tmp-write at `assets.py:230-239`. Rejection never touches disk — tests assert `os.listdir(tmp_path) == []`.
- Scope is provider-fetched assets only (Mouser / DigiKey allow-listed hosts). User uploads (which might carry exotic SVG/AVIF/HEIC) are NOT affected — they flow through a different code path. Provider CDNs serve standard PNG/JPEG/WebP/GIF/PDF, so no legitimate-asset-rejection risk.

**Workspace isolation (asset routes)**
- `GET /api/parts/assets/{ws_id}/{filename}` at `parts_assets.py:73-74` enforces `ws_id != ws.id` ⇒ 404. Path-traversal guard at `:76-77` intact. Unchanged by this PR.

**Tests**
- Inverted previously-permissive test → `…_unknown_body_with_known_ext_rejected`, asserts no on-disk write.
- New parametrised `test_magic_bytes_unknown_body_rejected_per_ext` covers all 5 allow-listed extensions.
- `_resp()` fixture default body now carries valid PNG magic so existing accept-path tests stay green under the stricter rule. Clean refactor.

**CI**: backend-tests / lockfile-drift / pip-audit / playwright-e2e / prod-validate / web-build all green; deploy skipping. Pre-merge gate: 0 fail/pending. MERGEABLE.

Merging.

Issue #253, opened 2026-05-02
[CQ-005 follow-up] Lint gates non-blocking; no flip-to-blocking deadline


matescb:
## Verification pass — 2026-05-02

`ruff check app --exit-zero` and `npm run lint || true` ship today (`.github/workflows/ci.yml:119` and `:185`). Both configs say "flip off in a follow-up PR." No follow-up scheduled.

## Suggested fix

Two-step:

1. Baseline current violations:
   ```bash
   ruff check app --output-format=github > .ruff-baseline.txt
   npm run lint --workspace web -- --output-file .eslint-baseline.txt
   ```
2. Wrap CI gates so they fail only on **new** violations beyond the baseline. Or commit to triaging the baseline within N weeks and dropping `--exit-zero`/`|| true`.

## Verify

Pick one date, write it in the issue, and hold the team to it.

matescb:
## Implementation plan

### Context

`.github/workflows/ci.yml` ships two non-blocking lint gates today:

- `ruff check app --exit-zero` (line 64). Comment at lines 59-61 says
  "first land is non-blocking; flip off in a follow-up PR once the
  existing violations are triaged."
- `npm run lint || true` (line 92). Comment at lines 87-89 says the
  same.

The issue notes the line numbers in the report (119 and 185) are
stale; CI today is 173 lines and the gates sit at 64 and 92. No
follow-up PR has been opened. Both `--exit-zero` and `|| true` are
load-bearing only because nobody has triaged the baseline.

Repo posture from CLAUDE.md: "There is no Python linter and no JS
linter configured. CI's only static check is `tsc -b`...". This is
half-true today — ruff and eslint *are* invoked, but they don't gate
the build, so functionally CLAUDE.md is right. The presence of
non-blocking gates is the worst of both worlds: they slow CI without
preventing regressions.

### Approach

Two-step, exactly as the issue suggests:

1. **Baseline.** Generate a deterministic snapshot of current ruff
   and eslint violations into checked-in files. Future runs subtract
   the baseline and fail only on the delta. This unblocks gating
   without forcing an immediate triage of every existing lint.

2. **Optional flip-day.** Pick a date (default: 6 weeks from merge),
   write it into the comment block in `ci.yml`, and triage the
   baseline by then. On flip-day, drop the baseline files and the
   `--exit-zero` / `|| true` suffixes.

The "fail only on new violations" wrap can be done two ways:

[...]

PR #296, opened 2026-05-03
fix(api): import datetime in builds + orders activity routes

### LINKED ISSUE #279: [CRITICAL] `builds.py` and `orders.py` activity-cursor endpoints crash with NameError — `datetime` not imported

Two activity-cursor endpoints will 500 the moment a real cursor is parsed. `ruff --select F821` proves it.

## Evidence

`backend/app/api/routes/builds.py:191`:
```python
cursor_at = datetime.fromisoformat(before_occurred_at)
```

`backend/app/api/routes/orders.py:319`:
```python
cursor_at = datetime.fromisoformat(before_occurred_at)
```

Neither file has `from datetime import datetime` (or any datetime import). Confirmed by:
```bash
$ grep -nE "^(from datetime|import datetime)" backend/app/api/routes/builds.py backend/app/api/routes/orders.py
(no output)
```

The first time a client passes `?cursor=…` to either endpoint, the call hits this line and raises `NameError: name 'datetime' is not defined` → 500 to the user.

## Why CI didn't catch it

This is exactly what `ruff --select F821` exists to catch. CI runs ruff but with `--exit-zero` (issue #253), so any F821 is reported but not failing. This finding is the case-in-point for flipping the gate to blocking.

## Suggested fix

Add the import to both files:

```python
from datetime import datetime
```

Then run `ruff check backend/app --select F821` locally and confirm zero hits.

## Verify

- Add a regression test that hits both endpoints with a valid `?cursor=…` value (encode/round-trip a real cursor) and asserts 200.
- After flipping ruff to blocking, this regression class becomes impossible to ship.

## Cross-refs

- #253 — lint gate non-blocking (this is the consequence)
- #251 — `raise_http()` migration (touches the same routers; perfect time to also add the import sweep)

matescb:
## Summary

Both `backend/app/api/routes/builds.py` and `backend/app/api/routes/orders.py` call `datetime.fromisoformat(before_occurred_at)` to parse the activity cursor query param, but neither file imports `datetime`. The first time a client passes `?before_occurred_at=…`, the call raises `NameError: name 'datetime' is not defined` and the endpoint returns 500.

The local-variable annotation `cursor_at: datetime | None` does not crash because both files have `from __future__ import annotations`, so the annotation is a string — only the runtime call was reachable.

This is exactly the class of bug `ruff --select F821` exists to catch (cross-ref #253 — lint gate currently non-blocking).

## Changes

- `backend/app/api/routes/builds.py`: add `from datetime import datetime`
- `backend/app/api/routes/orders.py`: add `from datetime import datetime`
- `backend/tests/test_activity_cursor_builds_orders.py`: new regression suite — 4 tests covering valid cursor (200) and malformed cursor (422) for both endpoints

## Test plan

- [x] `pytest tests/test_activity_cursor_builds_orders.py` — 4/4 pass
- [x] `pytest tests/test_builds.py tests/test_orders.py tests/test_activity_pagination.py` — 32/32 pass (no regressions)
- [x] `ruff check backend/app/api/routes/{builds,orders}.py --select F` — clean
- [x] Full backend suite — only pre-existing unrelated flakes (`test_pagination.py` / `test_parts_router_split.py` — both fail on schema-setup / route-registration unrelated to this change)

Closes #279

### DIFF
--- a/backend/app/api/routes/builds.py
+++ b/backend/app/api/routes/builds.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from datetime import datetime
 from uuid import UUID
 
 from fastapi import APIRouter, HTTPException, Query, Request, status
--- a/backend/app/api/routes/orders.py
+++ b/backend/app/api/routes/orders.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import logging
+from datetime import datetime
 from uuid import UUID
 
 from fastapi import APIRouter, HTTPException, Query, Request, status
--- a/backend/tests/test_activity_cursor_builds_orders.py
+++ b/backend/tests/test_activity_cursor_builds_orders.py
@@ -1,0 +1,135 @@
+"""Regression test for issue #279.
+
+`builds.py` and `orders.py` both used `datetime.fromisoformat(...)` in their
+activity-cursor endpoints without importing `datetime`. The first request
+with a `?before_occurred_at=…` parameter would 500 with NameError. These
+tests exercise the cursor-parsing path end-to-end so the missing import is
+caught at runtime.
+"""
+from __future__ import annotations
+
+import uuid
+
+import pytest
+from fastapi.testclient import TestClient
+
+from app.main import app
+from tests._factories import (
+    create_part as _create_part,
+    create_project_with_bom as _create_project_with_bom,
+    create_storage as _create_storage,
+    signup_user,
+)
+
+
+@pytest.fixture
+def authed():
+    c = TestClient(app)
+    signup_user(c)
+    return c
+
+
+def test_build_activity_cursor_param_accepted(authed):
+    """GET /api/builds/{id}/activity with a valid cursor returns 200, not 500."""
+    c = authed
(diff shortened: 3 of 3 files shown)


matescb:
**claude-review: approved (trivial fix-and-merge)**

Diff is exactly what the title says — adds `from datetime import datetime` to `backend/app/api/routes/builds.py` and `orders.py`, plus a 135-line regression suite covering the four cursor paths (valid + malformed × builds + orders). Verified the PR's single commit `d032540` matches the 3-file / +137 / -0 `gh pr view --files` summary — no rider changes.

Bug class confirmed: `cursor_at: datetime | None` annotation survived only via `from __future__ import annotations` (string-evaluated); the runtime `datetime.fromisoformat(before_occurred_at)` call NameError'd 500 on every client passing the cursor. Tests now exercise that exact path so a future regression is caught.

CI green across the board (backend-tests, web-build, playwright-e2e, prod-validate, pip-audit, lockfile-drift, line-count-budget); MERGEABLE; not a fork. No `CLAUDE.md` invariant touched.

Merging `--squash --delete-branch`.

PR #297, opened 2026-05-03
fix(security): refuse to boot prod without SMTP; never log verification link

### LINKED ISSUE #281: [security] Email verification leaks full verification links to logs/stdout when SMTP is unconfigured in prod

Prod email verification is forced on at signup, but `docker-compose.prod.yml` doesn't pass SMTP credentials or `APP_BASE_URL`. The mailer falls back to stdout — every signup writes the verification link to container logs.

## Evidence

`backend/app/core/mail.py:69`:
```python
def _smtp_configured() -> bool:
    s = settings()
    return bool(s.SMTP_HOST and s.SMTP_USER and s.SMTP_PASSWORD)
```

`backend/app/core/mail.py:77` (approx — fallback path):
```python
if not _smtp_configured():
    log.info("dev-mode email (no SMTP configured): %s", body)
    return
```

`backend/app/core/config.py:132`:
```python
SMTP_HOST: str = ""
SMTP_USER: str = ""
SMTP_PASSWORD: str = ""
APP_BASE_URL: str = "http://localhost:5173"
```

`docker-compose.prod.yml:72` — backend service env block — does NOT pass `SMTP_HOST`, `SMTP_USER`, `SMTP_PASSWORD`, or `APP_BASE_URL`.

Net result in prod:
1. User signs up.
2. Backend generates a verification token.
3. Mailer sees `_smtp_configured() == False`, writes the body (containing the verification URL) to stdout.
4. URL lands in `docker compose logs backend`, persists per the json-file driver's retention (10MB × 5 files).
5. Anyone with `docker logs` access can claim the account by visiting the link.

Worse: if `APP_BASE_URL` defaults to localhost in prod, the link is also wrong — but the token itself is real.

## Suggested fix

**(a)** Fail closed in prod — refuse to mount the email-verification flow if `APP_ENV == "prod"` and SMTP is unconfigured. Mirror the `WORKSPACE_SECRETS_KEY` validator pattern in `core/config.py`.

**(b)** Add `SMTP_HOST`, `SMTP_USER`, `SMTP_PASSWORD`, `APP_BASE_URL` to `docker-compose.prod.yml` backend env (and `deploy/.env.prod.example`).

**(c)** Even with SMTP configured, never log the body of an outbound email. The dev-mode log line should be a one-time warning ("SMTP not configured — emails will be discarded; set SMTP_* env"), not the message body.

Recommend doing all three. (a) prevents the failure mode in prod; (b) wires the production path; (c) hardens the dev fallback.

## Verify

- After fix, a prod-shaped startup with empty `SMTP_*` should refuse to boot (analogous to `WORKSPACE_SECRETS_KEY`).
- `grep "verification" /var/log/.../backend.log` returns no hits even when SMTP is misconfigured (the message is dropped, not logged).
- A new test in `backend/tests/test_mail.py`: when `_smtp_configured()` is False, `send_email(...)` does not write the body to any log handler.

## Cross-refs

- `8fe347b` — original commit that landed email verification.
- #26 — `WORKSPACE_SECRETS_KEY` fail-closed pattern is the template.

matescb:
## Summary

- Add `_require_smtp_in_prod` Settings validator (mirrors the existing `_require_workspace_secrets_key_in_prod` pattern). Refuses to boot in prod when any of `SMTP_HOST` / `SMTP_USER` / `SMTP_PASSWORD` / `MAIL_FROM` / `APP_BASE_URL` is empty (or `APP_BASE_URL` is the dev default).
- Drop the `cfg.SMTP_HOST` half of the dispatch in `send_verification_email` so prod always selects the SMTP backend; the stdout backend now refuses to run in prod (defence in depth) and logs only a single WARNING in dev (no more multi-line `print` banner).
- Wire `SMTP_HOST` / `SMTP_PORT` / `SMTP_USER` / `SMTP_PASSWORD` / `MAIL_FROM` / `APP_BASE_URL` into the `docker-compose.prod.yml` backend env block (no `:-` defaults on required values, so a missing entry surfaces both as a Compose warning and as the new validator's hard fail). Document the new requirements in `deploy/.env.prod.example`.

## Why

Issue #281: prod email verification was forced on at signup, but the backend silently fell through to a stdout mail backend whenever any `SMTP_*` var was empty — writing the verification token URL into container logs where anyone with `docker logs` access could claim a fresh signup. Three-part remediation per the issue's implementation plan: fail closed in prod, wire the env, harden the dev fallback.

## Test plan

- [x] New `backend/tests/test_mail_prod_safety.py` (7 tests):
  - Validator rejects prod with empty SMTP and with the dev default `APP_BASE_URL`.
  - Validator boots prod cleanly with full SMTP + public `APP_BASE_URL`.
  - Dev posture preserved (empty SMTP env allowed).
  - `_send_stdout` raises `RuntimeError` in prod and produces no log/stdout containing the verification token (regression test for the leak).
  - `send_verification_email` selector goes to SMTP backend in prod regardless of `SMTP_HOST`.
  - Dev `_send_stdout` logs the link at WARNING and does not print to stdout.
- [x] Updated `test_prod_with_valid_workspace_secrets_key_boots` to also set the SMTP vars now required by the new validator.
- [x] `pytest -q` full backend suite: 738 passed, 2 skipped, 3 deselected.

## Deploy notes

Operator must add real values for `SMTP_HOST` / `SMTP_USER` / `SMTP_PASSWORD` / `MAIL_FROM` / `APP_BASE_URL` to `.env.prod` on the VPS before approving the production deploy environment in GitHub. The deploy will fail fast (backend container exits with `ValidationError`) if any required field is empty — that's the intended failure mode, not a regression. Verify with `docker compose -f docker-compose.prod.yml --env-file .env.prod config | grep SMTP` before approving.

Closes #281

### DIFF
--- a/backend/app/core/config.py
+++ b/backend/app/core/config.py
@@ -74,14 +74,19 @@ class Settings(BaseSettings):
     # Email / SMTP for the email-verification flow (SEC2-014).
     # In dev (APP_ENV != "prod") the mail backend writes to stdout so the
     # verification link surfaces in container logs without an SMTP server.
-    # In prod all four SMTP_* vars must be set.
+    # In prod the model_validator below rejects any of SMTP_HOST / SMTP_USER /
+    # SMTP_PASSWORD / MAIL_FROM / APP_BASE_URL being empty (or APP_BASE_URL
+    # being the dev default), so a misconfigured deploy fails fast at boot
+    # instead of silently falling through to the stdout backend and writing
+    # the verification link to container logs (issue #281).
     SMTP_HOST: str = ""
     SMTP_PORT: int = 587
     SMTP_USER: str = ""
     SMTP_PASSWORD: str = ""
     MAIL_FROM: str = "noreply@stockmanager.local"
     # Public-facing base URL for generating verification links.
-    # In dev the default mirrors the Vite dev server.
+    # In dev the default mirrors the Vite dev server; the prod validator
+    # rejects this default (links must be on the public hostname).
     APP_BASE_URL: str = "http://localhost:5173"
     # Enable the two-step email-verification flow on signup (SEC2-014).
     # Defaults to True in prod, False elsewhere so the test suite can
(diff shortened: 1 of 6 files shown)


matescb:
Approved by claude-review (no medium+ findings, CI green). Verified: prod-gated SMTP requirement matches WORKSPACE_SECRETS_KEY pattern; dev still boots with empty SMTP; `_send_stdout` raises RuntimeError in prod before any logger/print of the link; SMTP backend never logs the body; `_scrub_event` drops cookies/auth/non-GET request bodies from Sentry; `test_prod_with_empty_smtp_fails_closed` pins the boot-fail assertion.

PR #299, opened 2026-05-03
fix(stock): enforce storage constraints on orders.receive and builds.consume output

### LINKED ISSUE #280: [BE-004 follow-up] `orders.receive` and `builds.consume` output bypass storage constraints

BE-004 (#35) was fixed for `add_stock` and `move_stock` via `_enforce_storage_constraints` in `backend/app/domain/stock/service.py:333-388`. But two other producer paths write `StockEntry` directly without that guard — same partial-close shape as BE-005 vs `match_entry`.

## Evidence

`backend/app/domain/orders/service.py:38` — `receive()` line write:
```python
entry = StockEntry(
    workspace_id=workspace_id,
    part_id=oe.part_id,
    lot_id=lot.id,
    storage_location_id=line.storage_location_id,
    ...
)
```
No call to `_enforce_storage_constraints`. So receive can land stock into a `single_part_only` storage that already holds another part, or into a storage that has `existing_parts_only=True` for a part with no prior entries.

`backend/app/domain/builds/service.py:437` — `consume()` output_entry:
```python
output_entry = StockEntry(
    workspace_id=workspace_id,
    part_id=sub_part.id,
    lot_id=output_lot.id,
    storage_location_id=storage.id if storage else None,
    quantity_delta=build.quantity,
    ...
)
```
Same — sub-assembly output bypasses the storage constraints.

The `_enforce_storage_constraints` helper is private to `stock/service.py`. Either it should be promoted to a callable that orders/builds also use, or both ledger-write paths should be funnelled through `add_stock`.

## Suggested fix

Two shapes:

**(a)** Hoist `_enforce_storage_constraints` out of `stock/service.py`'s private API and call it explicitly from `orders.receive` and `builds.consume` before each `StockEntry(...)` write.

**(b)** Refactor receive/build-output to call `add_stock(...)` through the public API (which already enforces constraints + locks). This is cleaner but requires the input shapes to match — may need a thin adapter.

(a) is lower-risk and ships in one PR.

## Verify

Extend `backend/tests/test_storage_constraints.py` with two cases:
- A `single_part_only=True` storage holds part A. `POST /api/orders/<id>/receive` for part B targeting that storage → expect 4xx, not silent success.
- A `single_part_only=True` storage. Build with `associated_subassembly_part_id` for part C, `output_storage_location_id` = same storage → expect 4xx.

## Cross-refs

- #35 — original BE-004 (closed; only patched add/move).
- #242 — BE-005 had the same partial-close shape (`match_entry` fixed, `create_order`/`add_entry` weren't).

matescb:
## Summary

BE-004 (#35) added `single_part_only` / `existing_parts_only` enforcement via `_enforce_storage_constraints` in `backend/app/domain/stock/service.py`, but only `add_stock` and `move_stock`'s destination side called it. Two other producer paths wrote `StockEntry` rows directly and bypassed the guard:

- `orders.receive` (`domain/orders/service.py`) — a receive line could land stock into a `single_part_only` bin already holding another part, or into an `existing_parts_only` bin where the receiving part has no prior positive history.
- `builds.consume` sub-assembly output (`domain/builds/service.py`) — a build whose project has `associated_subassembly_part_id` could deposit the produce-side row into a constrained bin.

This PR (shape (a) from the issue):

- Renames `_enforce_storage_constraints` -> `enforce_storage_constraints` (public) in `domain/stock/service.py` and updates the existing two internal call-sites. No back-compat alias needed (no external callers existed).
- Calls `enforce_storage_constraints` from `orders.receive` (per receive line, after the storage workspace/archived/full checks) and from `builds.consume` (for the sub-assembly output, when an `output_storage_location_id` is supplied). Both call sites already hold the per-(workspace, part) advisory lock via `lock_parts_for_stock_write`, so the helper's nested per-(workspace, storage) lock composes without deadlock.
- Adds `except StockConflictError` arms in `routes/orders.py::receive_order` and `routes/builds.py::consume_build` mapping the violation to a structured 409 with the same `{message, constraint, storage_location_id}` body shape used by `/api/stock/add` (mirror of `routes/stock.py`). Frontend handlers that already parse this shape work unchanged.

Hard invariants preserved: workspace isolation (no new cross-table joins; helper queries are workspace-filtered), append-only ledger (no schema change), API envelope (`{data, status}` via `responses.ok` / detail-spread `HTTPException`).

## Test plan

- [x] `tests/test_storage_constraints.py` extended with four new cases (all green):
  - `test_orders_receive_rejects_single_part_only_bin_holding_other_part` — 409 with `constraint=single_part_only`.
  - `test_orders_receive_rejects_existing_parts_only_for_unstocked_part` — 409 with `constraint=existing_parts_only`.
  - `test_builds_consume_output_rejects_single_part_only_bin` — 409 on the build-output write.
  - `test_builds_consume_output_allows_unconstrained_bin_regression` — sanity that the new guard does not fire for plain bins.
- [x] `pytest tests/test_storage_constraints.py tests/test_orders.py tests/test_builds.py` -> 38 passed locally against Postgres.
- [ ] Manual smoke in dev: stock part A in a `single_part_only` bin, post `/api/orders/{id}/receive` for part B targeting the same bin -> expect 409 JSON body with `constraint`.

Closes #280
Refs #35

### DIFF
--- a/backend/app/api/routes/builds.py
+++ b/backend/app/api/routes/builds.py
@@ -21,6 +21,7 @@ from app.domain.builds.service import (
 )
 from app.domain.projects.models import Project
 from app.domain.stock.models import StockEntry
+from app.domain.stock.service import StockConflictError
 
 router = APIRouter()
 
@@ -165,6 +166,19 @@ def consume_build(
     try:
         result = consume(
             db, workspace_id=ws.id, user_id=user.id, build=b, project=project, payload=payload
+        )
+    except StockConflictError as exc:
+        # BE-004 follow-up (#280): the build sub-assembly output is a
+        # producer write into the chosen output storage; if that storage
+        # is constrained the violation surfaces as a structured 409 with
+        # the same body shape as /api/stock/add.
+        raise HTTPException(
+            status_code=status.HTTP_409_CONFLICT,
+            detail={
+                "message": str(exc),
+                "constraint": exc.constraint,
+                "storage_location_id": str(exc.storage_location_id),
+            },
         )
     except BuildError as exc:
         # `get_db` rolls back on raise (BE2-010), so dropping the
--- a/backend/app/api/routes/orders.py
+++ b/backend/app/api/routes/orders.py
@@ -22,6 +22,7 @@ from app.domain.orders.schemas import (
 from app.domain.orders.service import OrderError, receive
 from app.domain.parts.models import Part
 from app.domain.stock.models import StockEntry
+from app.domain.stock.service import StockConflictError
 
 router = APIRouter()
 logger = logging.getLogger(__name__)
@@ -295,6 +296,18 @@ def receive_order(order_id: UUID, payload: ReceiveIn, db: DbSession, ws: Current
     o = _get_order(db, ws.id, order_id)
     try:
         result = receive(db, workspace_id=ws.id, user_id=user.id, order=o, payload=payload)
+    except StockConflictError as exc:
+        # BE-004 follow-up (#280): receive lines into a constrained
+        # destination must surface a structured 409 with the same body
+        # shape as /api/stock/add so existing client-side handlers work.
+        raise HTTPException(
+            status_code=status.HTTP_409_CONFLICT,
+            detail={
+                "message": str(exc),
+                "constraint": exc.constraint,
+                "storage_location_id": str(exc.storage_location_id),
+            },
+        )
     except OrderError as exc:
         # Re-raise as a 4xx — `get_db` rolls back automatically when
         # the route raises (BE2-010), so we don't need an explicit
--- a/backend/app/domain/builds/service.py
+++ b/backend/app/domain/builds/service.py
@@ -23,7 +23,11 @@ from app.domain.lots.models import Lot
 from app.domain.parts.models import Part, PartMetaMember, PartSubstitute
 from app.domain.projects.models import Project, ProjectEntry
 from app.domain.stock.models import StockEntry
-from app.domain.stock.service import current_quantity, lock_parts_for_stock_write
+from app.domain.stock.service import (
+    current_quantity,
+    enforce_storage_constraints,
+    lock_parts_for_stock_write,
+)
 from app.domain.storage.models import StorageLocation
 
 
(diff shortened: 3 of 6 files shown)


matescb:
**claude-review (Opus 4.7)**: APPROVE

Verified the diff extends the per-(workspace, storage) advisory lock pattern from PR #223 to `orders.receive` and `builds.consume` output writes. Lock key shape is correct (per-(ws, storage), not per-(ws, part)). Lock ordering is part-then-storage in every caller; `lock_parts_for_stock_write` already sorts part UUIDs. `StockConflictError` catch is positioned before the broader domain-error catch in both routes — correct ordering. Workspace isolation preserved (helper queries are ws-filtered; storage objects are belongs-checked before the helper call). API envelope shape mirrors `/api/stock/add`. Hard invariants (append-only ledger, no `inventory.qty`, MPN uniqueness) untouched.

Four new constraint-enforcement tests cover the orders-receive and builds-consume positive/negative cases. CI green: 8/8 real checks pass.

**Non-blocking nits** (worth a follow-up):

1. `_lock_for_storage_constraint` docstring at L129–132 still claims "the multi-part helpers — `builds.consume`, `orders.receive` — do not pass a destination through `enforce_storage_constraints`, so they don't take a storage lock." This is now stale; both helpers now do.

2. Multi-storage receive deadlock potential. Within one `orders.receive` transaction, multiple receive lines can land into multiple distinct storages, and the per-storage locks are taken in payload-line order (not sorted UUID). Two concurrent receives on different orders that share storage destinations in inverse order can deadlock. Postgres will detect (40P01) and abort, surfacing as a 500 — not silent corruption — but it would be cleaner to pre-sort distinct destination storages and acquire all per-storage locks up front, mirroring `lock_parts_for_stock_write`. (`builds.consume` is unaffected — only one output storage per call.) The order-row `with_for_update()` serialises same-order receives, so the practical exposure is cross-order receives only. Low likelihood, worth tightening when convenient.

3. The new tests are sequential constraint checks, not concurrent regression tests for the producer-side races this PR claims to close. The pre-existing `test_single_part_only_cross_part_race_blocked` only covers `/api/stock/add`. A threaded receive-vs-receive (or receive-vs-add) test into the same `single_part_only` bin would exercise the new lock surface.

None of these block merge.

PR #301, opened 2026-05-03
fix(security): stream provider asset downloads to enforce size cap (#285)

### LINKED ISSUE #285: [security] Provider asset size cap is checked AFTER fully buffering response into memory

A malicious or compromised provider can return a multi-GB body. The size cap (`_MAX_BYTES`) is enforced on `len(body)` AFTER `httpx` has buffered the entire response. Memory pressure / OOM before any rejection happens.

## Evidence

`backend/app/domain/parts/services/assets.py:169` — `_http_get`:
```python
resp = client.get(url, follow_redirects=False)
```
No `stream=True`. httpx reads the whole body into `resp.content`.

`backend/app/domain/parts/services/assets.py:206`:
```python
body = resp.content
if not body or len(body) > _MAX_BYTES:
    return None
```

So a 500MB datasheet response is fully buffered before the size check fires — the rejection is correct, but the OOM has already happened.

## Suggested fix

Stream the response and abort early:

```python
with client.stream("GET", url, follow_redirects=False) as resp:
    if resp.status_code != 200:
        return None
    chunks: list[bytes] = []
    total = 0
    for chunk in resp.iter_bytes(chunk_size=64 * 1024):
        total += len(chunk)
        if total > _MAX_BYTES:
            log.warning("provider asset rejected: exceeds %d bytes from %s", _MAX_BYTES, url)
            return None
        chunks.append(chunk)
    body = b"".join(chunks)
```

Bonus: combine with a `Content-Length` pre-check — many providers send the header, so we can reject before opening the body stream at all.

## Verify

- Add a test that mocks `httpx.Response.iter_bytes` to yield a stream exceeding `_MAX_BYTES`; assert `fetch_provider_asset` returns `None` AND that no chunks beyond the cap are accumulated.
- Manual: serve a 100MB body from a local mock; assert backend RSS doesn't grow by 100MB.

## Cross-refs

- SEC2-006 — original SSRF/host-allow-list (the body-cap check landed in the same area).
- #246 — magic-byte validation gap (same file).

matescb:
## Summary

The `_MAX_BYTES = 10 MB` ceiling on provider image / datasheet downloads in `backend/app/domain/parts/services/assets.py` was enforced on `len(resp.content)` **after** `httpx.Client.get()` had already buffered the full response into memory. A hostile or compromised provider CDN could ship a multi-GB body and OOM the backend long before the rejection fired.

This PR rewrites `_http_get` to stream the body via `httpx.Client.stream()` and abort as soon as the running chunk total exceeds `_MAX_BYTES`, so an oversize hostile body is never fully accumulated.

## Changes

- `backend/app/domain/parts/services/assets.py`
  - Add `_AssetResponse(status_code, headers, body)` dataclass; `body=None` signals an oversize abort.
  - Rewrite `_http_get` to use `client.stream("GET", url)` with chunked reads (64 KB) and a running-total guard that bails out immediately past `_MAX_BYTES`.
  - Belt-and-braces: short-circuit on a declared `Content-Length > _MAX_BYTES` before reading any body bytes (treated as advisory — a missing or malformed header falls through to the chunk-counting guard).
  - Update `fetch_provider_asset` call-site: read `resp.body` (and treat `None` as a refusal) instead of `resp.content`. The redundant `len(body) > _MAX_BYTES` post-check is removed because the cap is now enforced upstream.
  - `_ext_from_response` now takes a `headers: dict` instead of an `httpx.Response`.
- `backend/tests/test_assets.py`, `backend/tests/test_provider_assets.py`
  - Update `_resp` helpers to return `_AssetResponse` instead of a `MagicMock` shaped like `httpx.Response`.
  - Update `test_fetch_returns_none_on_oversize` to set `body=None` (the new oversize signal).
  - Add three streaming-path tests:
    - `test_http_get_streams_and_aborts_when_exceeds_max_bytes` — asserts `_http_get` abandons the iterator after the 2nd chunk and never reads the 3rd (which would have crossed the cap by ~1 KB).
    - `test_http_get_aborts_on_content_length_pre_check` — asserts a `Content-Length: _MAX_BYTES+1` short-circuits before `iter_bytes` is called at all.
    - `test_http_get_ignores_malformed_content_length` — asserts a junk `Content-Length` value (`"banana"`) doesn't crash and the body is still read.

## Invariants preserved

- Magic-byte validation from #246 (strict: sniffed must equal declared) — unchanged and still applied to the streamed body.
- Host allow-list + DNS-resolves-to-public-IP SSRF guard (SEC2-006) — unchanged.
- `follow_redirects=False` — unchanged.
- No `verify=False` on httpx clients.
- Content-addressed storage path `{UPLOAD_DIR}/parts/{ws_id}/{sha}.{ext}` — unchanged.
- API envelope, workspace isolation — not touched.

## Test plan

- [x] `pytest tests/test_provider_assets.py tests/test_assets.py -q` — 35 passed locally (Postgres-backed).
- [x] `lint-delta` — no new ruff violations (baseline 159 -> 155, 4 fixed since baseline; nothing added).
- [x] Manual scenarios covered by the new tests:
  - Stream that crosses the cap mid-download → aborts; body=None.
  - `Content-Length` header that declares oversize → short-circuit, no body read.
  - Malformed `Content-Length` → falls through to chunk guard.
- [ ] Production smoke after merge: confirm a normal Mouser/DigiKey image still downloads and gets content-addressed under `/uploads/parts/{ws}/{sha}.png`.

Closes #285

### DIFF
--- a/backend/app/domain/parts/services/assets.py
+++ b/backend/app/domain/parts/services/assets.py
@@ -33,6 +33,7 @@ import ipaddress
 import logging
 import os
 import socket
+from dataclasses import dataclass
 from urllib.parse import urlparse
 
 import httpx
@@ -44,6 +45,24 @@ log = logging.getLogger(__name__)
 
 _TIMEOUT_SEC = 10.0
 _MAX_BYTES = 10 * 1024 * 1024  # 10 MB ceiling — datasheet PDFs are usually 1-3 MB
+_CHUNK_SIZE = 64 * 1024  # 64 KB — streaming read granularity for the size-cap guard
+
+
+@dataclass
+class _AssetResponse:
+    """Result of `_http_get` after size-capped streaming.
+
+    `body` is `None` when the download was aborted because either the
+    `Content-Length` header or the running stream total exceeded
+    `_MAX_BYTES`. Returning the headers + status separately lets the
+    caller distinguish "oversize" from "non-200" / "empty" without
+    holding any oversized bytes in memory.
+    """
+
+    status_code: int
+    headers: dict[str, str]
+    body: bytes | None
+
 
 # Content-Type → file extension. Falls through to URL-suffix inference
 # for anything that doesn't match. SVG is intentionally absent — see
@@ -117,8 +136,8 @@ def _ext_from_url(url: str) -> str | None:
     return ext if 1 <= len(ext) <= 5 and ext.isalnum() else None
 
 
-def _ext_from_response(resp: httpx.Response, url: str) -> str:
-    ct = (resp.headers.get("content-type") or "").split(";", 1)[0].strip().lower()
+def _ext_from_response(headers: dict[str, str], url: str) -> str:
+    ct = (headers.get("content-type") or "").split(";", 1)[0].strip().lower()
     if ct == "image/svg+xml":
         # Explicit refusal — the file lands as `.bin` and the serve
         # route forces an `attachment` disposition for non-images.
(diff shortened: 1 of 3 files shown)


matescb:
**claude-review: APPROVE & MERGE**

Streaming rewrite of `_http_get` correctly addresses #285. Verified critical concerns:

- **Cumulative byte cap enforced before buffering**: `total += len(chunk)` is checked against `_MAX_BYTES` BEFORE `chunks.append(chunk)`, so a hostile multi-GB upstream is bounded at ~10 MB + one 64 KB chunk in memory before refusal. ✓
- **Per-chunk read timeout**: `httpx.Client(timeout=_TIMEOUT_SEC)` propagates the 10s timeout to the streaming read path; a slow-drip upstream hits `ReadTimeout` per chunk and `fetch_provider_asset`'s `except Exception` maps it to refusal. ✓
- **Magic-byte validation order preserved**: still runs in `fetch_provider_asset` on `body[:16]` after `_http_get` returns the fully-buffered (capped) body. ✓
- **Content-addressed path unchanged**: `{UPLOAD_DIR}/parts/{ws_id}/{sha}.{ext}` preserved. Oversize aborts return `None` before any disk I/O — no partial files. Pre-existing tempfile + `os.replace` rename logic is untouched. ✓
- **Cap = 10 MB**: aligned with #201/#263. ✓
- Content-Length pre-check is correctly advisory (try/except ValueError covers malformed values; test `test_http_get_ignores_malformed_content_length` proves the fallthrough).
- `follow_redirects=False` + host allow-list + DNS-globally-routable SSRF guard all preserved.

Tests are thorough: the streaming-abort test asserts the iterator is abandoned by chunk-counting (`chunks_consumed == [len(a), len(b)]`), proving chunk C is never read. CI green (8/8 real checks). Merging.

PR #311, opened 2026-05-03
fix(auth): preserve query string + hash on RedirectIfAuthed deep-link bounce

### LINKED ISSUE #304: [FE] Authenticated redirects drop query string + hash from deep links
## Verification pass — 2026-05-04 (round 4)

The `RedirectIfAuthed` guard preserves `pathname` only. A user clicking a deep link like `/parts/scan-import?storage_id=abc&tab=queue` (or with a hash anchor) gets bounced to login, then post-auth lands at `/parts/scan-import` without the query/hash — the original intent is silently lost.

## Evidence

`web/src/App.tsx:145`:

```tsx
function RedirectIfAuthed({ children }: { children: React.ReactNode }) {
  const { me, loading } = useAuth();
  const location = useLocation();
  if (loading) return null;
  if (me) {
    const from = (location.state as { from?: Location } | null)?.from;
    const target =
      from && from.pathname !== "/login" && from.pathname !== "/signup"
        ? from.pathname              // ← only pathname; search and hash dropped
        : "/parts";
    return <Navigate to={target} replace />;
  }
  return <>{children}</>;
}
```

Likely the same pattern at the login flow's complement (where `from` is captured) — verify both ends.

## Why this matters

Scan-import URLs, BOM filter URLs, deep-linked detail tabs (`/parts/abc?tab=specs`), and any future shareable link all lose intent on the auth bounce. Users blame "the app forgot what I clicked".

## Suggested fix

Build the navigation target from all three components:

```tsx
const target =
  from && from.pathname !== "/login" && from.pathname !== "/signup"
    ? from.pathname + (from.search ?? "") + (from.hash ?? "")
    : "/parts";
return <Navigate to={target} replace />;
```

Or use the `Location` shape directly if `Navigate` accepts it (`<Navigate to={from} replace />` works in react-router v6 if `from` is a partial Location).

Audit the symmetric site too — wherever `state.from = location` is set on the way INTO the login flow, ensure `location` is the full Location object (not just pathname).

## Verify

- Add an RTL test: visit `/parts/scan-import?storage_id=abc#row-7` while logged out → bounce to `/login` → log in → expect `Navigate` target equals the original full URL.
- Manual: log out, paste a query-string URL, log in, confirm the query string survives.

## Cross-refs

- FE2-010 (closed) — original deep-link replay; this is a refinement.

matescb:
## Summary

- `RedirectIfAuthed` in `web/src/App.tsx` was sending already-authed users to `from.pathname` only, dropping `search` and `hash`. Deep links like `/parts/scan-import?storage_id=abc&tab=queue` or `/parts/abc?tab=specs#anchor` silently lost intent on the auth round-trip.
- Pass the full `Partial<Path>` (`pathname` + `search` + `hash`) to `<Navigate to={…}>`. Also imports `Location` as a type from `react-router-dom` (the old code accidentally relied on the global DOM `Location` because both share `pathname`).
- The complementary code paths (`Gate`, `authBus` 401 handler, `Login.tsx`, `Signup.tsx`) already preserved the full Location — no change needed there.

## Test plan

- [x] `npm test` — 238 pass / 3 skipped, including the existing 5 `RedirectIfAuthed` tests and the new `state.from with search+hash` test that asserts query string and fragment survive the redirect.
- [x] `npm run build` — `tsc -b` clean, vite build succeeds.
- [ ] Manual smoke: log out, paste `/parts/scan-import?storage_id=abc&tab=queue` into the address bar, log in, confirm the URL after redirect still has `?storage_id=abc&tab=queue`.
- [ ] Manual smoke: same with `#row-7` fragment.

Closes #304

### DIFF
--- a/web/src/App.tsx
+++ b/web/src/App.tsx
@@ -1,5 +1,5 @@
 import { lazy, Suspense, type ReactNode } from "react";
-import { Navigate, Outlet, Route, Routes, useLocation } from "react-router-dom";
+import { Navigate, Outlet, Route, Routes, useLocation, type Location } from "react-router-dom";
 import { AuthProvider, useAuth } from "@/lib/auth";
 import AppShell from "@/components/layout/AppShell";
 import { ConfirmDialogProvider } from "@/components/ConfirmDialog";
@@ -143,11 +143,17 @@ function RedirectIfAuthed({ children }: { children: React.ReactNode }) {
   if (loading) return null;
   if (me) {
     const from = (location.state as { from?: Location } | null)?.from;
-    const target =
-      from && from.pathname !== "/login" && from.pathname !== "/signup"
-        ? from.pathname
-        : "/parts";
-    return <Navigate to={target} replace />;
+    // Preserve search + hash so deep-links like /parts/scan-import?storage_id=abc
+    // or /parts/abc?tab=specs#anchor survive the auth round-trip (#304).
+    if (from && from.pathname !== "/login" && from.pathname !== "/signup") {
+      return (
+        <Navigate
+          to={{ pathname: from.pathname, search: from.search, hash: from.hash }}
+          replace
+        />
+      );
+    }
+    return <Navigate to="/parts" replace />;
   }
   return <>{children}</>;
 }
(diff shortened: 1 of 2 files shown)


matescb:
Approved by claude-review (no medium+ findings, CI green). Verified: redirect target is built from the in-app `state.from` Location only — no `?next=` URL parsing, no open-redirect surface. `<Navigate to={{pathname, search, hash}}>` is a relative SPA navigation. Test asserts both query string and fragment survive the bounce. /accept-invitation flow is unrelated (token POST in settings, no route guard).

PR #312, opened 2026-05-03
fix(tests): default conftest DATABASE_URL to 127.0.0.1 for host pytest

### LINKED ISSUE #306: [testing] `conftest.py` defaults DATABASE_URL to docker hostname — host pytest fails on DNS
## Verification pass — 2026-05-04 (round 4)

Outside Docker, the backend test suite fails before any test executes — `conftest.py` defaults `DATABASE_URL` to the docker network hostname (`db:5432`). The CLAUDE.md instructions claim local pytest works via `TEST_DATABASE_URL` override, but the default-fallback always points at a hostname that isn't resolvable on the host.

## Evidence

`backend/tests/conftest.py:17`:

```python
os.environ.setdefault(
    "DATABASE_URL",
    os.environ.get("TEST_DATABASE_URL", "postgresql+psycopg://stockmgr:stockmgr@db:5432/stockmgr_test"),
)
```

`backend/tests/conftest.py:94` — engine creation that fails on DNS:

```python
eng = create_engine(settings().DATABASE_URL, future=True)
try:
    with eng.connect():
        pass
except Exception:
    admin_url = settings().DATABASE_URL.rsplit("/", 1)[0] + "/postgres"
    ...
```

Reproduction:
```bash
$ cd backend && pytest -q tests/test_provider_assets.py::test_magic_bytes_match_pdf_accepted
collected 1 item
...
sqlalchemy.exc.OperationalError: (psycopg.OperationalError) connection failed:
  could not translate host name "db" to address: ...
```

Even a unit-style test that doesn't touch the DB fails because the engine fixture is session-autouse.

## Why this matters

CLAUDE.md says local pytest is supported. In practice it isn't unless the operator pre-exports `TEST_DATABASE_URL` AND knows what value to use. New contributors waste 15-30 minutes diagnosing this before reading the conftest carefully.

## Suggested fix

Two options:

**(a)** Default to `127.0.0.1:5432` (the local-postgres-listening shape) instead of `db:5432`. The Docker compose setup explicitly overrides `DATABASE_URL` per-service so the change doesn't break in-container tests.

**(b)** Detect docker vs host: if `socket.gethostbyname("db")` raises, fall back to `127.0.0.1`. More magic but zero-config.

(a) is simpler. Update the docstring + CLAUDE.md note: "outside Docker, ensure local Postgres listens on 127.0.0.1:5432 with database `stockmgr_test`."

## Verify

```bash
# Without TEST_DATABASE_URL set:
unset TEST_DATABASE_URL DATABASE_URL
cd backend && pytest -q tests/test_provider_assets.py
```

Should pass against a local Postgres (assuming one is running) without environment-variable gymnastics.

## Cross-refs

- TEST-009 (closed) — savepoint isolation; perf fix, but didn't address discoverability.
- CLAUDE.md "Tests" section — claims this workflow works.

matescb:
## Summary

- `backend/tests/conftest.py` defaulted `DATABASE_URL` to the docker-network hostname `db:5432`, which only resolves inside the compose network. Host-side `pytest` failed at DNS resolution before any test ran unless the operator already knew to export `TEST_DATABASE_URL`.
- This PR flips the fallback to `127.0.0.1:5432` so a stock local Postgres setup (user/password `stockmgr`/`stockmgr`, `CREATEDB` privilege) just works with `cd backend && python -m pytest -q`.
- In-container runs are unchanged: `docker-compose.dev.yml` injects `DATABASE_URL=...@db:5432/...` for the `backend` service before Python starts, so `os.environ.setdefault` is a no-op and the docker hostname is preserved.
- The same now-dead-but-consistent fallback in `tests/test_migrations.py` is updated to match.
- `CLAUDE.md` "Tests" section documents the zero-config host flow plus the `TEST_DATABASE_URL` override path for non-default setups.

## Test plan

- [x] Verify `127.0.0.1` resolves and is reachable on the dev host.
- [x] Confirm `os.environ.setdefault` no-ops when `DATABASE_URL` is pre-set (in-container path preserved verbatim).
- [x] Confirm `TEST_DATABASE_URL` override still wins over the new default.
- [ ] CI: backend `pytest` passes against the CI Postgres (no DNS regression in-container).

Closes #306

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -45,13 +45,26 @@ cd web && npm test                                                           # v
 ```
 
 Outside Docker, the backend test suite needs a real Postgres (no SQLite
-fallback — schema uses `UUID`, `ARRAY`, `Numeric`):
+fallback — schema uses `UUID`, `ARRAY`, `Numeric`). `tests/conftest.py`
+defaults `DATABASE_URL` to
+`postgresql+psycopg://stockmgr:stockmgr@127.0.0.1:5432/stockmgr_test`, so
+with a local Postgres listening on `127.0.0.1:5432` and the `stockmgr` user
+(password `stockmgr`, `CREATEDB` privilege) you can simply run:
+
+```bash
+cd backend && python -m pytest -q
+```
+
+The test DB (`stockmgr_test`) is auto-created by conftest if it doesn't
+exist. For a non-default host/port/credentials, set `TEST_DATABASE_URL`:
 
 ```bash
 TEST_DATABASE_URL=postgresql+psycopg://stockmgr:stockmgr@127.0.0.1:5432/stockmgr_test \
   python -m pytest -q
 ```
 
+Note: the dev compose file does not publish Postgres to the host, so host
+pytest needs a separately installed local Postgres (not the Docker one).
 `tests/conftest.py` drops + recreates the public schema between tests.
 
 ### Build / lint
--- a/backend/tests/conftest.py
+++ b/backend/tests/conftest.py
@@ -14,8 +14,17 @@ from sqlalchemy.orm import Session, sessionmaker
 # This exercises the explicit-DATABASE_URL-override branch in
 # Settings._assemble_database_url (INFRA2-005): when DATABASE_URL is
 # supplied directly it is used as-is and the POSTGRES_* parts are ignored.
+#
+# The fallback default points at 127.0.0.1 so host-side `pytest` works
+# against a local Postgres without any environment override (issue #306).
+# In-container runs (`docker compose exec backend pytest`) are unaffected:
+# Compose injects DATABASE_URL=...@db:5432/... before Python starts, so
+# `setdefault` is a no-op and the docker hostname is preserved. Operators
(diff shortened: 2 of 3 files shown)


matescb:
**claude-review: APPROVED**

Trivial DX fix. Verified:
- In-Docker runs preserved: `docker-compose.dev.yml:23` injects `DATABASE_URL=...@db:5432/...` into the backend container before Python starts, so `os.environ.setdefault` no-ops and the docker hostname survives.
- Override precedence intact: `TEST_DATABASE_URL` (then `DATABASE_URL` if pre-set) still wins over the new fallback.
- Schema drop+recreate (`DROP SCHEMA public CASCADE` / `CREATE SCHEMA public`) untouched in conftest.
- `tests/test_migrations.py` fallback updated for consistency (rarely hit since conftest sets `DATABASE_URL` first).
- CLAUDE.md doc update matches the new default and still documents the `TEST_DATABASE_URL` override path.

All non-meta checks green. Merging.

Issue #328, opened 2026-05-08
TP-108: Sourcing request budget tracker (parts-count)


matescb:
## Project Guardrails (read before coding — applies to every TP-* issue)

You are working in a multi-tenant FastAPI + React + Postgres app. Auto-deploy lands on prod the moment your PR merges to `main` (gated only by a human reviewer). Treat hygiene as part of the contract.

**Read first**
- [`CLAUDE.md`](../blob/main/CLAUDE.md) — hard invariants ("No `inventory.qty`", workspace isolation, API envelope, MPN uniqueness, content-addressed assets, `bag_signature`, provider catalog vs spec keys, **no `verify=False`**, "Things that have bitten us — don't undo").
- [`docs/ARCHITECTURE.md`](../blob/main/docs/ARCHITECTURE.md) — domain decomposition, ledger model, workspace isolation, API envelope, migrations, frontend conventions.
- [`docs/STYLE.md`](../blob/main/docs/STYLE.md) — every doc page (ADRs, phases, runbooks, API/domain/frontend reference) must conform. Read before adding any `*.md`.
- [`docs/development.md`](../blob/main/docs/development.md) — local dev loop, lint baselines.

**Non-negotiable invariants this issue must respect**
1. **API envelope** — every response uses `{ data, status }` via `responses.ok()` / `responses.err()`. Never bare payloads. Frontend HTTP goes through `web/src/lib/api.ts`.
2. **Workspace isolation in code** — every query against a workspace-owned table filters by `ws.id`; every cross-table FK from a request body/path is validated via `assert_in_workspace` (`backend/app/api/_helpers.py:22-49`). New routes must be reviewed by the `workspace-isolation-checker` agent and pinned with a test in `tests/test_workspace_isolation.py` style.
3. **Migrations** — append-only numbering (`NNNN_short_name.py`); reversible `downgrade()`; reviewed by `alembic-migration-reviewer`. Don't edit a merged migration — add a new one.
4. **No `verify=False`** on httpx clients (CI greps under `backend/app/`). Annotate with `# noqa: tls-verify` only for intentional test doubles.
5. **Integer-only quantities** (DB-005). Round MOQ overbuy up before persisting.
6. **Encrypted credentials** — provider/sourcing creds always go through `app.core.secrets.encrypt`/`decrypt`; plaintext never reaches a column or a serializer (`has_*: bool` masks on read).
[...]

PR #335, opened 2026-05-08
TP-108: Sourcing request budget tracker

matescb:
## Summary
- Adds the in-process TrustedParts parts-count budget tracker.
- Tracks live/degraded/blocked verdicts across 10s, 1m, 1h, and 24h windows with an injectable clock.
- Adds targeted tests for soft/hard thresholds, parts-count accounting, window expiry, workspace isolation, and singleton identity.

## Validation
- `uv run --extra dev pytest -k sourcing_budget`: passed, 7 passed
- `uv run --extra dev ruff check app/domain/sourcing/budget.py tests/test_sourcing_budget.py`: passed
- Docker validation unavailable: docker is not installed in this environment

Counters increment by `parts_count`, not by HTTP call count.

Refs #328

### DIFF
--- a/backend/app/domain/sourcing/budget.py
+++ b/backend/app/domain/sourcing/budget.py
@@ -1,1 +1,110 @@
-"""Request budget tracking scaffold for TrustedParts sourcing."""
+"""Per-workspace TrustedParts request budget tracking."""
+
+from __future__ import annotations
+
+import time
+from dataclasses import dataclass
+from typing import Callable, Literal
+from uuid import UUID
+
+
+@dataclass(frozen=True)
+class BudgetVerdict:
+    allow: bool
+    mode: Literal["live", "degraded", "blocked"]
+    reason: str
+
+
+# Soft thresholds: (rolling window seconds, parts searched in window).
+SOFT_LIMITS = [(10, 50), (60, 150), (3600, 2000), (86400, 15000)]
+
+# Hard thresholds: (rolling window seconds, parts searched in window).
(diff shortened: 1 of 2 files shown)


matescb:
**claude-review: APPROVE**

TP-108 budget tracker scaffold is sound:

- Per-workspace keying via `(workspace_id, window_seconds)` — workspace isolation verified by `test_two_workspaces_independent`.
- Counter increments by `parts_count` (not call count), matching the upstream pricing/quota model — verified by `test_counter_increments_by_parts_count_not_one`.
- Hard limits return `blocked` (allow=False), soft limits return `degraded` (allow=True), defaults to `live`. 4 rolling windows (10s/1m/1h/24h).
- Injectable monotonic clock; `_prune` correctly drops events older than the longest applicable window.
- Process-local singleton, explicitly tied to uvicorn `--workers 1` in the docstring with a TODO for a future Redis ADR if workers scale up — matches the project invariant in CLAUDE.md.
- No DB columns, no migration, no request-path wiring yet — pure scaffold module + tests. This sidesteps any TOCTOU/`FOR UPDATE` concerns for now.

Followups (not blockers, file as separate work):
1. When this is wired into `TrustedPartsClient`, the `check` → call → `record` sequence needs a lock (`threading.Lock` or `asyncio.Lock`) to be race-free. CPython GIL only guarantees atomic primitives, not the compound operation.
2. Tests poke `tracker._prune` directly — fine for unit scope but the public API path (`check` calls `_prune` internally) already exercises it.
3. No persistence: a backend restart resets the budget. Documented and acceptable for this stage.

Green CI (10/10), no security or workspace-isolation issues. Merging.

Issue #345, opened 2026-05-08
TP-202: POST /api/parts/{id}/sourcing/refresh


matescb:
## Project Guardrails (read before coding — applies to every TP-* issue)

You are working in a multi-tenant FastAPI + React + Postgres app. Auto-deploy lands on prod the moment your PR merges to `main` (gated only by a human reviewer). Treat hygiene as part of the contract.

**Read first**
- [`CLAUDE.md`](../blob/main/CLAUDE.md) — hard invariants ("No `inventory.qty`", workspace isolation, API envelope, MPN uniqueness, content-addressed assets, `bag_signature`, provider catalog vs spec keys, **no `verify=False`**, "Things that have bitten us — don't undo").
- [`docs/ARCHITECTURE.md`](../blob/main/docs/ARCHITECTURE.md) — domain decomposition, ledger model, workspace isolation, API envelope, migrations, frontend conventions.
- [`docs/STYLE.md`](../blob/main/docs/STYLE.md) — every doc page (ADRs, phases, runbooks, API/domain/frontend reference) must conform. Read before adding any `*.md`.
- [`docs/development.md`](../blob/main/docs/development.md) — local dev loop, lint baselines.

**Non-negotiable invariants this issue must respect**
1. **API envelope** — every response uses `{ data, status }` via `responses.ok()` / `responses.err()`. Never bare payloads. Frontend HTTP goes through `web/src/lib/api.ts`.
2. **Workspace isolation in code** — every query against a workspace-owned table filters by `ws.id`; every cross-table FK from a request body/path is validated via `assert_in_workspace` (`backend/app/api/_helpers.py:22-49`). New routes must be reviewed by the `workspace-isolation-checker` agent and pinned with a test in `tests/test_workspace_isolation.py` style.
3. **Migrations** — append-only numbering (`NNNN_short_name.py`); reversible `downgrade()`; reviewed by `alembic-migration-reviewer`. Don't edit a merged migration — add a new one.
4. **No `verify=False`** on httpx clients (CI greps under `backend/app/`). Annotate with `# noqa: tls-verify` only for intentional test doubles.
5. **Integer-only quantities** (DB-005). Round MOQ overbuy up before persisting.
6. **Encrypted credentials** — provider/sourcing creds always go through `app.core.secrets.encrypt`/`decrypt`; plaintext never reaches a column or a serializer (`has_*: bool` masks on read).
[...]

PR #362, opened 2026-05-08
TP-202: POST /api/parts/{id}/sourcing/refresh

matescb:
## Summary
- Added member-gated `POST /api/parts/{part_id}/sourcing/refresh` with a 6/min per-workspace SlowAPI limit, workspace-scoped part lookup, no-MPN 422 envelope, and the same sourcing exception mapping as the existing part read route.
- Added explicit `force_refresh` support in the sourcing service/cache path so refresh bypasses valid cache rows while still upserting the existing `(workspace_id, query_hash)` row and recording live budget use.
- Added focused route tests and updated the sourcing API docs plus the `/api/parts` route snapshot.

## Validation
- `docker compose -f docker-compose.dev.yml exec backend pytest -k sourcing_part_refresh_route` could not run: `docker: command not found`.
- `docker compose -f docker-compose.dev.yml exec backend ruff check` could not run: `docker: command not found`.
- `cd backend && uv run --extra dev python -m pytest -k sourcing_part_refresh_route` -> 7 passed, 830 deselected, 1 warning.
- `cd backend && uv run --extra dev python -m pytest tests/test_parts_routing.py -q` -> 1 passed, 1 warning.
- `cd backend && uv run --extra dev ruff check app/api/routes/sourcing.py app/domain/sourcing/cache.py app/domain/sourcing/service.py tests/test_sourcing_part_refresh_route.py tests/test_parts_routing.py` -> All checks passed.
- `cd backend && LINT_CMD="uv run --extra dev ruff check app --output-format=concise" BASELINE_FILE="../.ruff-baseline.txt" WORK_DIR="." bash ../scripts/lint-delta.sh` -> no new violations (baseline: 159, current: 154, fixed since baseline: 5).
- `cd backend && uv run --extra dev ruff check` was also attempted as the closest full Ruff command; it fails on existing repository baseline violations, so lint-delta is the meaningful gate outside Docker.

## ttl_seconds=0 vs force_refresh
- Used a new `force_refresh: bool = False` kwarg instead of `ttl_seconds=0`.
- Reason: the existing cache helper checks for a valid cache row before TTL affects writes, and `ttl_seconds=0` would write an immediately expiring replacement instead of the required fresh row with the normal 30-minute expiry.

## Deviations
- Docker validation was unavailable in this environment because the `docker` executable is not installed.
- Updated `backend/tests/test_parts_routing.py` in addition to the named test file because this repo has a route snapshot that must include deliberate new `/api/parts` routes.

Refs #345

### DIFF
--- a/backend/app/api/routes/sourcing.py
+++ b/backend/app/api/routes/sourcing.py
@@ -211,6 +211,83 @@ def get_part_sourcing(
     )
 
 
+@parts_router.post(
+    "/{part_id}/sourcing/refresh",
+    dependencies=[Depends(require_role("member"))],
+)
+@limiter.limit("6/minute", key_func=workspace_key)
+def refresh_part_sourcing(
+    request: Request,
+    part_id: UUID,
+    ws: CurrentWorkspace,
+    user: CurrentUser,
+    db: Session = Depends(get_db),
+):
+    part = assert_in_workspace(db, Part, part_id, ws.id, label="part")
+    mpn = (part.mpn or "").strip()
+    if not mpn:
+        return _error_response(request, 422, "validation_error", "part has no MPN")
+
+    try:
+        out = sourcing_service.search(
+            db,
+            workspace=ws,
+            mpns=[mpn],
+            use_cached_data=False,
+            ttl_seconds=_PART_SOURCING_TTL_SECONDS,
+            requested_by=user.id,
+            force_refresh=True,
+        )
+    except SourcingNotConfigured:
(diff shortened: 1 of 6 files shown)


matescb:
Approved by claude-review (no medium+ findings, CI green). Workspace isolation via assert_in_workspace, 6/min/ws rate limit (tighter than GET's 60/min, matches connection-test pattern), BUDGET.record on every refresh (force_refresh guarantees cache miss), cache upsert on (ws_id, query_hash) prevents row duplication, identical 4-category error mapping + envelope shape as TP-201.

PR #389, opened 2026-05-08
TP-507: Replenishment cost report (no persistent history)

matescb:
## Summary
- Added transient `GET /api/reports/replenishment-cost` with a reports-domain service/schema, workspace-scoped stock aggregation via stock helpers, 4-hour sourcing cache TTL, Decimal cost math, currency mismatch reasons, and graceful sourcing status instead of sourcing 409/503 responses.
- Added the Replenishment Cost report UI, route, sort control, TrustedParts attribution/source labels, status banner, totals table, and Vitest coverage.
- Documented the API/frontend behavior and added ADR-0020 section `Reports policy: no persistent price history`.

## Validation
- `docker compose -f docker-compose.dev.yml exec backend pytest -k replenishment_cost_report` not run: Docker is unavailable in this environment (`docker: command not found`).
- `cd backend && uv sync --frozen --extra dev`
- `cd backend && uv run python -m pytest -q -k replenishment_cost_report` -> `8 passed, 884 deselected, 1 warning`
- `cd backend && uv run ruff check app/domain/reports tests/test_replenishment_cost_report.py` -> `All checks passed!`
- `cd backend && uv run ruff check --select I,F401,F821 app/api/routes/reports.py` -> `All checks passed!`
- `cd web && npm install` -> completed with Node 18 engine warnings for transitive packages and existing `npm audit` moderate findings
- `cd web && npm run typecheck` -> passed
- `cd web && npm test -- ReplenishmentCostReport` -> `1 passed (1), 3 passed (3)`
- `cd web && npx eslint src/routes/reports/ReplenishmentCostReport.tsx src/routes/reports/__tests__/ReplenishmentCostReport.test.tsx` -> passed
- `git -C /mnt/data/WORK/stockManager-1/.codex/worktrees/tp-507-382 diff --check` -> passed

## No persistent history confirmation
No migration, table, or column was added for TrustedParts-derived replenishment/price history. The integration assertion is:

```python
assert not any(
    fragment in name
    for name in schema_names
    for fragment in forbidden_fragments
)
```

## ADR section
ADR-0020 now includes `Reports policy: no persistent price history`, documenting that replenishment-cost prices are recomputed from existing lot costs plus the short-lived sourcing cache on each request: `docs/adr/0020-trustedparts-sourcing-provider-split.md`.

## Deviations / CI risk
- Full `cd backend && uv run ruff check` is still blocked by existing repository lint baseline violations outside this PR; scoped Ruff checks for new/touched report code pass.
[...]

PR #387, opened 2026-05-08
TP-506: BOM buyability report

matescb:
## Summary
- Added backend BOM buyability report schema/service and `GET /api/reports/bom-buyability` with workspace-scoped active-project selection, 50-project cap/truncated flag, `build_quantity >= 1`, rate limit, cache-default true, 4h Source-BOM cache TTL, and graceful sourcing status fallback.
- Added React BOM buyability report page with URL-persisted build quantity, DataTable rows, status/truncation badges, TrustedParts attribution/source label, and Source-BOM deep links.
- Added backend/frontend tests and API/frontend docs.

## Validation
- `TEST_DATABASE_URL=postgresql+psycopg://stockmgr:stockmgr@127.0.0.1:5432/stockmgr_test_tp506 uv run --extra dev python -m pytest -q tests/test_bom_buyability_report.py` -> 7 passed, 2 warnings.
- `npm test -- BomBuyabilityReport` -> 1 file passed, 3 tests passed.
- `npm run typecheck` -> passed.
- `npx eslint src/routes/reports/BomBuyabilityReport.tsx src/routes/reports/__tests__/BomBuyabilityReport.test.tsx src/App.tsx src/routes/reports/Reports.tsx` -> passed.
- `LINT_CMD="ruff check app --output-format=concise" BASELINE_FILE=".ruff-baseline.txt" WORK_DIR="backend" bash scripts/lint-delta.sh` -> no new violations.
- `LINT_CMD="npm run lint -- --format=unix" BASELINE_FILE="../.eslint-baseline.txt" WORK_DIR="." bash ../scripts/lint-delta.sh` from `web/` -> no new violations.
- `git -C /mnt/data/WORK/stockManager-1/.codex/worktrees/tp-506-381 diff --check` -> passed.

## Screenshot
- Not captured: this environment has no Docker/dev backend available for an authenticated populated browser session.

## Deviations / Limitations
- Docker validation commands could not run because `docker` is not installed in the environment.
- Raw `npm run lint` still reports existing baseline issues in unrelated files (`bagCode.ts`, `ZxingScanner.tsx`, etc.); scoped ESLint for touched frontend files and the baseline-aware lint check both pass.
- Raw app-wide `ruff check app` reports existing baseline issues outside this change; scoped ruff for new backend files and the baseline-aware ruff check pass.

Refs #381

PR #386, opened 2026-05-08
TP-505: Low-stock report with sourcing column set

matescb:
## Summary
- Adds `include_sourcing=true` to the low-stock report with 4-hour TrustedParts cache TTL, workspace dashboard cache preference, per-row sourcing data, and graceful `sourcing_status` flags.
- Adds low-stock sourcing columns, URL-persisted toggle, TrustedParts attribution, status banners, and per-row `CreateOrderLineModal` draft-PO action.
- Adds backend and frontend tests plus report API/frontend docs and provider-outage runbook note.

## Validation
- `docker compose -f docker-compose.dev.yml ps` could not run: Docker is not installed in this environment.
- `cd backend && uv run python -m pytest -q tests/test_reports.py tests/test_low_stock_report_with_sourcing.py` -> 13 passed, 1 warning.
- `cd backend && uv run ruff check app/domain/reports/service.py tests/test_low_stock_report_with_sourcing.py app/domain/sourcing/schemas.py` -> passed.
- `cd web && npm run typecheck && npm test -- LowStockReport && npx eslint src/routes/reports/Reports.tsx src/routes/reports/__tests__/LowStockReport.test.tsx` -> passed; Vitest reported React Router future-flag warnings only.
- `cd web && npm run lint` still reports pre-existing baseline issues outside this PR (`ZxingScanner`, `bagCode`, responsive grid/order/parts/storage files); changed report files pass targeted ESLint.
- `git -C /mnt/data/WORK/stockManager-1/.codex/worktrees/tp-505-380 diff --check` -> passed.

## Screenshot Note
- Not captured: Docker/dev server is unavailable in this environment, so I could not run the full app stack for toggle-off/toggle-on/budget-blocked screenshots.

## Deviations
- The existing low-stock implementation lived in `backend/app/api/routes/reports.py`, not a report service. This PR extracts only low-stock report logic into `backend/app/domain/reports/service.py` and leaves the route as a thin wrapper.
- Full raw backend `ruff check app` and web `npm run lint` are affected by existing repository lint baselines; scoped checks were run for changed files.

Refs #380

Issue #384, opened 2026-05-08
BOM-001: Opt-in auto-create missing parts on BOM import (quantity=0)


matescb:
## Project Guardrails (read before coding — applies to every TP-* issue)

You are working in a multi-tenant FastAPI + React + Postgres app. Auto-deploy lands on prod the moment your PR merges to `main` (gated only by a human reviewer). Treat hygiene as part of the contract.

**Read first**
- [`CLAUDE.md`](../blob/main/CLAUDE.md) — hard invariants ("No `inventory.qty`", workspace isolation, API envelope, MPN uniqueness, content-addressed assets, `bag_signature`, provider catalog vs spec keys, **no `verify=False`**, "Things that have bitten us — don't undo").
- [`docs/ARCHITECTURE.md`](../blob/main/docs/ARCHITECTURE.md) — domain decomposition, ledger model, workspace isolation, API envelope, migrations, frontend conventions.
- [`docs/STYLE.md`](../blob/main/docs/STYLE.md) — every doc page (ADRs, phases, runbooks, API/domain/frontend reference) must conform. Read before adding any `*.md`.
- [`docs/development.md`](../blob/main/docs/development.md) — local dev loop, lint baselines.

**Non-negotiable invariants this issue must respect**
1. **API envelope** — every response uses `{ data, status }` via `responses.ok()` / `responses.err()`. Never bare payloads. Frontend HTTP goes through `web/src/lib/api.ts`.
2. **Workspace isolation in code** — every query against a workspace-owned table filters by `ws.id`; every cross-table FK from a request body/path is validated via `assert_in_workspace` (`backend/app/api/_helpers.py:22-49`). New routes must be reviewed by the `workspace-isolation-checker` agent and pinned with a test in `tests/test_workspace_isolation.py` style.
3. **Migrations** — append-only numbering (`NNNN_short_name.py`); reversible `downgrade()`; reviewed by `alembic-migration-reviewer`. Don't edit a merged migration — add a new one.
4. **No `verify=False`** on httpx clients (CI greps under `backend/app/`). Annotate with `# noqa: tls-verify` only for intentional test doubles.
5. **Integer-only quantities** (DB-005). Round MOQ overbuy up before persisting.
6. **Encrypted credentials** — provider/sourcing creds always go through `app.core.secrets.encrypt`/`decrypt`; plaintext never reaches a column or a serializer (`has_*: bool` masks on read).
[...]

PR #390, opened 2026-05-08
BOM-001: Opt-in auto-create missing parts on BOM import

matescb:
## Summary

- Add `auto_create_missing_parts` to BOM import preview/commit schemas and return preview/commit counters for auto-created and skipped rows.
- Preserve the existing six-step match priority; only create a zero-stock local `Part` after no match is found, using the import workspace and `linked_provider='none'`.
- Skip auto-create rows with neither MPN nor part/name, merge same-MPN/name rows through flushed in-transaction parts, and surface `IntegrityError` as a rollback-triggering 409.
- Add focused BOM auto-create tests and update user, API, and domain docs.

## Validation

Docker validation was unavailable in this environment: `docker: command not found`. Closest local validation was run against local Postgres after `uv sync --frozen --extra dev`.

- `.venv/bin/python -m pytest -q tests/test_bom_import.py tests/test_bom_import_auto_create.py`
  - `17 passed, 3 warnings in 1.80s`
  - Existing `test_bom_import.py` behavior passes unchanged as part of this run.
- `.venv/bin/python -m pytest -q -k "bom_import or workspace_isolation"`
  - `68 passed, 851 deselected, 5 warnings in 18.33s`
- `.venv/bin/ruff check app/domain/projects/bom_import.py app/domain/projects/schemas.py tests/test_bom_import_auto_create.py`
  - `All checks passed!`
- `LINT_CMD="ruff check app --output-format=concise" BASELINE_FILE="../.ruff-baseline.txt" WORK_DIR="." bash ../scripts/lint-delta.sh`
  - `lint-delta: no new violations (baseline: 159, current: 149, fixed since baseline: 10)`
- `git diff --check`
  - no output

Workspace-isolation review: `_match_part` remains workspace-filtered, every inserted `Part` uses the import `workspace_id`, and `test_workspace_isolation_two_workspaces_same_mpn` covers same-MPN imports across workspaces.

## Sample CSV and Counters

```csv
qty,mpn,part
1,EXISTING-MPN,
2,NEW-MPN,
3,,Local only
4,,
```

Before, with `auto_create_missing_parts=false`:

```json
{ "inserted": 4, "matched": 1, "unmatched": 3, "auto_created": 0, "skipped": 0 }
```

After, with `auto_create_missing_parts=true`:

```json
{ "inserted": 3, "matched": 1, "unmatched": 0, "auto_created": 2, "skipped": 1 }
```

Preview with the same mapping reports:

```json
{ "would_auto_create_count": 2, "would_skip_count": 1 }
```

## Deviations

- Preview needed the existing route DB/workspace context to compute real match-aware counters, so the preview route now passes `db` and `workspace_id` into `bom.preview`; the API envelope remains unchanged.
- Unscoped local `uv run pytest` initially hit an Alembic import issue before `uv sync --frozen --extra dev`; the final local validation used `.venv/bin/python -m pytest`.

## Merge Order

No known merge-order dependency. I inspected open PRs #389, #387, #386, and #241 before publishing; their changed files do not overlap this PR's BOM import, projects API, or projects/BOM docs files.

Refs #384

### DIFF
--- a/backend/app/api/routes/projects.py
+++ b/backend/app/api/routes/projects.py
@@ -256,7 +256,7 @@ def del_entry(project_id: UUID, entry_id: UUID, db: DbSession, ws: CurrentWorksp
 @router.post("/{project_id}/bom/preview")
 def preview_bom(project_id: UUID, payload: BomImportPreviewIn, db: DbSession, ws: CurrentWorkspace):
     _get(db, ws.id, project_id)
-    return ok(bom.preview(payload).model_dump())
+    return ok(bom.preview(payload, db=db, workspace_id=ws.id).model_dump())
 
 
 @router.post("/{project_id}/bom/import")
--- a/backend/app/domain/projects/bom_import.py
+++ b/backend/app/domain/projects/bom_import.py
@@ -5,12 +5,12 @@ import base64
 import csv
 import io
 from dataclasses import dataclass
-from typing import Iterable
 from uuid import UUID
 
 import chardet
 from fastapi import status
-from sqlalchemy import and_, func, select
+from sqlalchemy import func, select
+from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm import Session
 
 from app.core.errors import ErrorCodes, raise_http
@@ -96,7 +96,16 @@ def _looks_numeric(s: str) -> bool:
         return False
 
 
-def preview(payload: BomImportPreviewIn) -> BomImportPreviewOut:
+class _SkipRow(Exception):
+    """Auto-create cannot materialise a usable Part from this row."""
+
+
+def preview(
+    payload: BomImportPreviewIn,
+    db: Session | None = None,
+    *,
+    workspace_id: UUID | None = None,
+) -> BomImportPreviewOut:
     raw = _decode_b64(payload.text_b64)
     enc = _detect_encoding(raw, payload.encoding)
     text = raw.decode(enc, errors="replace")
@@ -112,15 +121,27 @@ def preview(payload: BomImportPreviewIn) -> BomImportPreviewOut:
             headers=None,
             rows=[],
         )
-    has_header = payload.has_header if payload.has_header is not None else _looks_like_header(all_rows[0])
+    has_header = (
+        payload.has_header
+        if payload.has_header is not None
+        else _looks_like_header(all_rows[0])
+    )
     headers = all_rows[0] if has_header else None
     body = all_rows[1:] if has_header else all_rows
+    would_auto_create_count, would_skip_count = _preview_auto_create_counts(
+        payload,
+        body,
+        db=db,
+        workspace_id=workspace_id,
+    )
     return BomImportPreviewOut(
         detected_separator=sep,
         detected_encoding=enc,
         has_header=has_header,
         headers=headers,
         rows=[BomPreviewRow(cells=r) for r in body[:200]],
+        would_auto_create_count=would_auto_create_count,
+        would_skip_count=would_skip_count,
     )
 
 
@@ -172,7 +193,11 @@ def _split_designators(value: str, sep: str) -> list[str]:
     return [p for p in parts if p]
 
 
-def _apply_mapping(cells: list[str], mapping: list[BomMappingField], designator_sep: str) -> ParsedRow:
+def _apply_mapping(
+    cells: list[str],
+    mapping: list[BomMappingField],
+    designator_sep: str,
+) -> ParsedRow:
     out = ParsedRow()
     for m in mapping:
         idx = m.column_index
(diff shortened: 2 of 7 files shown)


matescb:
**claude-review: requesting changes**

Pipeline: `/review` + `/security-review` + `workspace-isolation-checker`. CI is green and `mergeable=MERGEABLE`, but one **high** finding blocks merge.

### High — `linked_provider="none"` breaks UI + behaves as provider-linked

`backend/app/domain/projects/bom_import.py::_auto_create_part` sets:

```python
linked_provider="none",
```

Two problems:

1. **Frontend Zod schema rejects it.** `web/src/lib/schemas.ts:59`:
   ```ts
   linked_provider: z.enum(["mouser", "digikey"]).nullable(),
   ```
   Any parts list / detail / search response that includes an auto-created part will throw `ZodError` on the client and break the page. None of the existing 1100 +/- shipped parts in prod ever had `linked_provider="none"` — the convention everywhere else (`parts_core.create_part`, `parts_scan`, `_parts_shared` serializer) is `None` for "not linked" and `"mouser"`/`"digikey"` when linked.

2. **Backend treats it as provider-linked.** `parts_core.py::patch_part:242`:
   ```python
   if p.linked_provider and not unlink:
       for f in ("manufacturer", "mpn"):
           ...HTTPException(... "is provider-owned on a linked part; pass unlink_provider=true to take ownership")
   ```
   With `"none"` (truthy), every PATCH that edits MPN or manufacturer on an auto-created part will 400 unless the user passes `unlink_provider=true` — directly contradicting this PR's user-doc claim that "Provider [is] set to `none`, so you can enrich it later." Same applies to the `description_locally_edited` flag at line 254 (`p.linked_provider is not None` flips it on for any description edit).

**Fix (one line):** drop `linked_provider="none"` (default `None` from the model column), or set it explicitly to `None`. The frontend doc string and `provider='none'` semantic intent should map to NULL on the column. With that change the part behaves identically to one created via `POST /api/parts` with no provider link — which is what the user-facing doc already promises.

A test guard would be nice: `assert part.linked_provider is None` in `test_auto_create_with_mpn` (currently asserts `linked_provider == "none"`, which encodes the bug).

### Other notes (non-blocking)

- **Idempotency** prompt-callout: I checked — the BOM commit endpoint has no idempotency key in this repo, so the "re-runs don't dup" concern is moot here. Re-import will simply re-match by MPN (priority 4 in `_match_part`) on the second run, so duplicates are prevented organically. Worth a note in `docs/api/projects.md` though.
- **MPN race → 409 (full rollback) rather than re-fetch + continue.** The `IntegrityError → raise 409` path is acceptable (not a 500), but if the spec calls for "race-resilient continue" you'd need a `try / db.flush() except IntegrityError: db.rollback() to savepoint; re-fetch by MPN`. Current behavior is conservative and matches `parts_core.create_part`'s pre-flight 409, so I'd leave it.
- **Audit-log entry** for auto-created parts: `parts_core.create_part` itself doesn't audit creation today (only archive/restore), so the BOM auto-create path matches existing behavior. If the BOM-001 spec wants traceability ("which import created which parts"), that's a separate small follow-up to add `_audit_log(action="part.created", ...)` in both code paths together.
- **Workspace isolation** ✓: `_match_part` and `_auto_create_part` both scope by `workspace_id`. `test_workspace_isolation_two_workspaces_same_mpn` covers same-MPN across workspaces correctly.
- **MPN uniqueness** ✓: `uq_parts_ws_mpn` is respected — the auto-create path uses `db.flush()` and the `IntegrityError → 409` handler covers the in-import race.
- **Opt-in default OFF** ✓: `auto_create_missing_parts: bool = Field(default=False)` on both schemas. Existing callers unaffected.

### What I want to see for re-review

1. `linked_provider=None` (or omit from the `Part(...)` constructor entirely).
2. Update `test_auto_create_with_mpn` (and any sibling tests that touch `linked_provider`) to assert `is None`.
3. Consider a tiny doc fix in `docs/user/projects-and-bom.md` — "Provider set to `none`" → "Provider unset (no link)" — to match what's actually persisted.

Marking `reopened`. Ping me on the next push.

matescb:
Addressed claude-review feedback in ae7d92a: auto-created BOM parts now persist linked_provider as NULL, the regression test asserts part.linked_provider is None, and the API/user docs now say provider is unset rather than provider='none'. Local checks: uv run ruff check app/domain/projects/bom_import.py tests/test_bom_import_auto_create.py; uv run python -m pytest -q tests/test_bom_import_auto_create.py; git diff --check.

matescb:
Approved by claude-review (sweep #100 fix verified: linked_provider=None now persists as NULL, Zod enum and PATCH guard both happy; CI green, no medium+ findings).

PR #317, opened 2026-05-03
docs: comprehensive documentation campaign

matescb:
## Summary

Five new doc shelves under `docs/`, retro-fills phases 01–03 + 11–13, expands runbooks 1→10, lands 18 ADRs, adds 18 in-tree module READMEs, deletes the legacy `docs/teardown/` audit set. Updates `CLAUDE.md` "Orientation" to point at the new tree.

**Final state**: 96 files / 12.9k lines under `docs/` (was 19 / 5.5k). 0 broken internal links across 347 internal references. Every new page carries an `Audience:` tag.

## What's new

| Shelf | Files | Audience | Subject |
|---|---:|---|---|
| `docs/api/` | 16 | engineer | per-router REST reference (15 routers + envelope/error/auth conventions) |
| `docs/domain/` | 11 | engineer | entity reference + ER diagram + invariants |
| `docs/frontend/` | 9 | engineer | routing, lib/api, TanStack patterns, components, scanner, testing |
| `docs/adr/` | 19 | engineer | 17 retro-ADRs from CLAUDE.md rules + ADR-0018 (SMTP fail-closed, surfaced by the audit pass) |
| `docs/runbooks/` | 11 | on-call | 10 ops runbooks + severity-matrix index (was just secret-rotation) |
| `docs/user/` | 12 | end user | task-shaped help pages — first-run, scan-import, BOM, builds, reports |
| `docs/phases/` | +6 | engineer | 01–03 retro-docs from migrations 0001–0005; 11–13 forward-docs from CHANGELOG |
| Module READMEs (in-tree) | 18 | engineer | per-module orientation under `backend/app/{domain,api/routes,core}/`, `web/src/{lib,components,routes}/` |

Plus `docs/README.md` (audience map) and `docs/STYLE.md` (style guide every page conforms to).

## Audited against `main`

The campaign was first executed against an older snapshot. After moving to a fresh branch off `origin/main`, three Explore agents audited the new docs against the 24 main-only commits since the original base. Patches applied:

- **#297** — refuse to boot prod without SMTP; never log verification link → new **ADR-0018**, plus `docs/api/auth.md`, `docs/runbooks/smtp-outage.md`
- **#299** — enforce storage constraints on `orders.receive` + `builds.consume` output → `docs/api/orders.md`, `docs/api/builds.md`, `docs/domain/orders-and-receive.md`, `docs/domain/builds-and-bom.md`
- **#298** — honour `?limit=N` on bare-list `/api/parts` → `docs/api/parts.md`
[...]

Issue #316, opened 2026-05-03
tags.py DELETE silently succeeds for cross-workspace IDs (inconsistent with rest of API)


matescb:
## What's wrong

Two delete routes in `backend/app/api/routes/tags.py` return success when the target id belongs to **another workspace**, instead of raising `404`:

- `DELETE /api/tags/{tag_id}`
- `DELETE /api/tags/links/{link_id}`

Most of the rest of the API raises `404 not_found` on cross-workspace lookups (e.g. `parts`, `orders`, `lots`, `storage`). Tags is the odd one out — it returns `200 OK` with no-op semantics.

## Why it matters

- **No data leak**: the other workspace's data is not exposed; the operation simply doesn't change anything.
- **API consistency**: clients can't distinguish "tag was deleted" from "tag belongs to a different workspace" by status code.
- **Workspace-isolation regression test gap**: `tests/test_workspace_isolation.py` currently passes for these routes precisely because they don't 404 — the isolation check passes, but the route's contract is silently weaker than the rest of the API.

## What needs deciding

Either:

1. **Fix to 404** — match the rest of the API. Add tests to `test_workspace_isolation.py` style. Probably right.
2. **Document as intentional** — add a `CLAUDE.md` "Things that have bitten us" entry and an ADR explaining why tags chose differently. Probably wrong, but worth surfacing.

## Where surfaced

Caught during the docs campaign while writing `docs/api/attachments-tags-cf.md` against `tags.py`.

Issue #315, opened 2026-05-03
audit_log model comment cites alembic migration 0024 (actual file is 0030)


matescb:
## What's wrong

`backend/app/domain/audit/models.py` has a model-level comment / docstring asserting that the `target_ids` GIN index was added in alembic migration `0024`. The 0024 migration **does not exist** — the chain skips from `0023_*` straight to `0025_*`. The audit_log table (and its GIN index) actually lands in `0030_audit_log.py`.

## Why it matters

Anyone using the model comment as a reference for "where did this index come from" will hit a dead end. Trivial fix; flagging because reviewing the migration chain is the primary way engineers reason about schema state.

## What needs doing

Update the comment in `backend/app/domain/audit/models.py` to cite `alembic/versions/0030_audit_log.py` instead of `0024`. Spot-check whether any other `0024`-citing comment / docstring exists elsewhere — `0024` was simply skipped in numbering, not silently squashed.

## Where surfaced

`docs/domain/data-model.md` (in the new `docs/comprehensive-doc-campaign` branch) noted the discrepancy while building the migration index.

Issue #314, opened 2026-05-03
CLAUDE.md cites non-existent backend/app/domain/parts/services/provider.py


matescb:
## What's wrong

`CLAUDE.md` "Hard invariants" — provider catalog vs spec keys section — says:

> Adding a new catalog field needs both sides.
> `web/src/lib/providerCatalog.ts` defines which custom-field keys are catalog metadata […]; the same key list lives server-side in `backend/app/domain/parts/services/provider.py`.

That backend file does **not** exist. `backend/app/domain/parts/services/` only contains `assets.py`, `bag_signature.py`, and `provider_cache.py`.

## Why it matters

The split between catalog metadata and user-curated specs is a hard invariant (see ADR-0007 in the new `docs/adr/` tree). Anyone trying to add a new catalog key by following the CLAUDE.md instruction will not find the file the doc points at, and may put the key in the wrong place — drifting the FE/BE lists. This is the exact failure mode the rule was written to prevent.

## What needs deciding

Either:

1. **Find where the canonical server-side catalog-key list actually lives** and update `CLAUDE.md` (and `docs/adr/0007-provider-catalog-vs-spec-split.md`, `docs/domain/providers.md`) to cite that path.
2. **Move the canonical list to `backend/app/domain/parts/services/provider.py`** (creating the file) so it matches what CLAUDE.md / ADR-0007 already document.

Option 2 is probably right — the `services/` directory is the natural home for cross-cutting parts logic, and the existing `provider_cache.py` lives there. A single `CATALOG_KEYS` set imported by both `bag_signature.py` and the catalog response builder would mirror `web/src/lib/providerCatalog.ts`.

## Where surfaced

Surfaced by two pages of the documentation campaign (PR with branch `docs/comprehensive-doc-campaign`):

- `docs/adr/0007-provider-catalog-vs-spec-split.md` — left a `TODO(verify)` against the missing path
- `docs/domain/providers.md` — same `TODO(verify)`

Both will need a follow-up edit once this is resolved.

PR #391, opened 2026-05-08
fix: clean up the three bugs PR #317's audit surfaced (#314, #315, #316)

matescb:
## Summary

Closes the three loose ends from PR #317's audit pass. Each was tracked as its own issue but never followed up by a PR until now.

- **#314** — CLAUDE.md "Provider catalog vs spec keys" pointed at a non-existent path. Updated the invariant to point at the real server-side touchpoints (`providers/base.py` for shapes, `parts_assets.py::_PROVIDER_RESERVED_KEYS` for the asset-side subset).
- **#315** — `audit/models.py` comment cited migration **0024** for the GIN index on `audit_log.target_ids`; actual file is `0030_audit_log.py`. One-character fix.
- **#316** — `DELETE /api/tags/links/{id}` returned 200 silently for foreign-workspace link ids. Every other DELETE returns 404 via `assert_in_workspace`; switched tags to the same. Added a regression test in `test_workspace_isolation.py` that pins both halves (B's DELETE on A's link 404s; A's link survives).

## Test plan

- [x] `pytest test_workspace_isolation.py::test_tag_unlink_rejects_foreign_workspace_id`
- [x] grep CLAUDE.md for the corrected paths exists
- [x] grep `audit/models.py` for "0030" comment

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -167,9 +167,11 @@ them, that's the bug.
 - **Provider catalog vs spec keys.** `web/src/lib/providerCatalog.ts`
   defines which custom-field keys are catalog metadata (price, stock,
   manufacturer URL, …) vs user-curated specs. The PartSpecs and
-  PartSourcing tabs split on this boundary; the same key list lives
-  server-side in `backend/app/domain/parts/services/provider.py`.
-  Adding a new catalog field needs both sides.
+  PartSourcing tabs split on this boundary; server-side, the catalog
+  field shapes live in `backend/app/domain/parts/providers/base.py`
+  and the asset-side reserved-keys subset is `_PROVIDER_RESERVED_KEYS`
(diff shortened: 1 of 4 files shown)


matescb:
CI fix pushed: corrected the new regression test to assert the surviving tag-link response shape (top-level id is link id; tag id is nested under tag.id).\n\nLocal verification:\n- uv run python -m pytest -q tests/test_workspace_isolation.py::test_tag_unlink_rejects_foreign_workspace_id\n- uv run python -m pytest -q tests/test_workspace_isolation.py\n- LINT_CMD="ruff check app --output-format=concise" BASELINE_FILE="../.ruff-baseline.txt" WORK_DIR="." bash ../scripts/lint-delta.sh\n- uv run python scripts/check_no_traceback_leaks.py\n- git diff --check\n\nMerge-order note: only #391 is open right now; no cross-PR merge-order dependency remains.

